PR #6, opened 2019-05-03
Minor changes for rand and  equals/hashcode

bnyra:
Hi @mboehm7,
as we discussed on Thursday, here is a minor PR for
- handling runtime seeds in lineage traces
- implementing equals() and hashCode() for `LineageItem`

Furthermore, I added some testcases.


mboehm7:
LGTM - thanks @bnyra. I only made some minor modifications of the hash computation and rand replacement during the merge.

PR #8, opened 2019-05-22
Required changes for time() builtin function.

phaniarnab:


mboehm7:
LGTM - thanks for the new builtin function. During the merge, I only made minor improvements (removed unnecessary imports, fixed formatting, and removed branches that were always true/false).

PR #9, opened 2019-06-04
Single node Read and Write for LIBSVM formatted files.

phaniarnab:
My branch is having 2 unnecessary commits, came with the 'pull' from origin/master. Not sure how to delete those entries or exclude from the PR.


mboehm7:
Thanks for the patch @phaniarnab. It was a good start but there were a number of issues, including some performance problems. I'll merge it together with my modifications, but I'd like to give you a summary:
* The handling of properties like header and sparse representation did not match the libsvm specs (and were also not used at all) - so I removed all related runtime integration.
* The computed ncol sizes were based on the number of non-zeros in the first row, which does not reflect the number of columns in terms of dimension size
* Write: The writers had unnecessary object allocation (stringbuilder per cell) in the critical path. Fixing this improved performance for 1M x 1K dense from 112s to 78s and for 1M x 1K sparse (sparsity=0.1) from 93s to 6s.
* Read: The readers parsed the entire sparse row for each and every cell, which lead to a O(N^2) parsing overhead per row. Fixing this improved performance for 100K x 1K from 1344s to 8s.
* For sparse reads, the class label append happened before the other appends, which means the resulting sparse row was not ordered and thus subsequent operations produced wrong results.
* All news files had the old ASF licenses not our Apache 2.0 license.

PR #10, opened 2019-06-07
[SYSTEMDS-73] Deduplication of nested for loops

bnyra:
This PR covers deduplication of nested for loops.
For loops in IfProgramBlocks are still not supported.
I also added some test cases too.


mboehm7:
LGTM - thanks for the patch @bnyra .

PR #11, opened 2019-06-11
[SYSTEMDS-70] Add lineage tracing for MultiReturnInstructions

bnyra:
I added lineage tracing for instruction with multiple output. E.g. `eigen()`

Therefore, I changed the method signature of `LineageTraceable` interface and adapt logic in `MultiReturnBuiltinCPInstruction`.


mboehm7:
LGTM - thank you so much @bnyra

PR #13, opened 2019-07-11
Pr parfor image ops

corepointer:
two parfor tests using image_brightness and image_mirror on a prepared mnist data set


mboehm7:
thanks @corepointer - that's a good start but please create the PR on a new branch without the commit of the previous PR, and replace the mnist dataset with synthetic data.

PR #15, opened 2019-07-12
parfor test on image operations mirror, brightness using synthetic data

corepointer:
One issue: brightness test seems to be inaccurate for reasons I could not uncover in a reasonable amount of time so I leave it as a FIXME for now to get the PR reviewed.
Set the eps value in ParForImageBrightnessTest.java to the usual accuracy to see the test fail.


mboehm7:
LGTM - thanks @corepointer. During the merge I made the following changes but there are still some correctness issues in `ParForImageBrightnessTest`:
* Replaced implementation of `parfor_image_mirror.R` with `apply(M, 2, rev)`
* Removed unnecessary imports in java test files
* Fixed the implementation of the parfor brightness test by missing as.integer casts in the dml script, and correct matrix reshaping by row in the R script.

PR #16, opened 2019-07-12
Add LargeDenseBlock Representations and support for Strings

kev-inn:
Adds a LargeDenseBlock representation for the DenseRowBlocks, while also fixing Bugs of the DenseRowBlocks.

Also adds a new DenseRowBlock for Strings (also a LDRB).

Tests were also added for LDRBs, while also adding some for not yet tested functionality.

Tests currently do not test LDRBs for large (greater than Integer.MAX_VALUE) array requirements.


docs/Tasks.txt
```diff
@@ -71,7 +71,8 @@ SYSTEMDS-100 Various Fixes
 SYSTEMDS-110 New Builtin Functions
  * 111 Time builtin function for script-level measurements            OK
  * 112 Image data augmentation builtin functions                      OK
- 
+ * 113 Builtin functions for linear regression algorithms             OK
```
mboehm7, line 75: Please, try to avoid leaking files across PRs.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlock.java
```diff
@@ -59,7 +59,23 @@ protected DenseBlock(int[] dims) {
...
+	 * @return the size of the dimension
+	 */
+	public final int getDim(int i) {
+	    if (i == 0) {
+	    	return _rlen;
+		}
+	    if (i == _odims.length) {
+	    	return _odims[i - 1];
+		}
+	    return _odims[i - 1] / _odims[i];
```
mboehm7: 1) The formatting tabs vs spaces is off
2) Maybe rewrite to single return and nested ternary operators (to improve readability)

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockBool.java
```diff
@@ -155,26 +155,31 @@ public DenseBlock set(int r, int c, double v) {
 	
 	@Override
 	public DenseBlock set(DenseBlock db) {
-		System.arraycopy(db.valuesAt(0), 0, _data, 0, _rlen*_odims[0]);
+		double[] data = db.valuesAt(0);
+	    for (int i = 0; i < _rlen*_odims[0]; i++) {
```
mboehm7: The formatting is off - please make a full pass over all modified classes in that regard as I'm not commenting on the remaining ones.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockBool.java
```diff
@@ -155,26 +155,31 @@ public DenseBlock set(int r, int c, double v) {
 	
 	@Override
 	public DenseBlock set(DenseBlock db) {
-		System.arraycopy(db.valuesAt(0), 0, _data, 0, _rlen*_odims[0]);
```
mboehm7, line 158: I would recommend to run a few experiments (maybe in a separate PR for performance improvements) and see if we could can something if the input DenseBlock is a DenseBlockBool (i.e., _data.clear(); _data.or(that._data))

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockFP32.java
```diff
@@ -149,7 +149,7 @@ public DenseBlock set(int r, int c, double v) {
 	
 	@Override
 	public DenseBlock set(DenseBlock db) {
-		System.arraycopy(db.valuesAt(0), 0, _data, 0, _rlen*_odims[0]);
+		System.arraycopy(DataConverter.toFloat(db.valuesAt(0)), 0, _data, 0, _rlen*_odims[0]);
```
mboehm7: Again maybe in a different PR (perf), I'd like to avoid unnecessary array copies (toFloat()) just to use an array copy after that. Probably the best would be to have a stream/loop to set all elements by using get double and a special case for common input/output arrays.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLBool.java
```diff
@@ -0,0 +1,366 @@
...
+		if (neededCompleteBlocks == _blocks.length) {
+			// Check if we can store enough rows in the last (most likely smaller) block
+			// and we don't need another row.
+			return neededLastBlockSize == 0 && possibleRowsPerBlock == lastBlockSize;
+		}
+		return false;
+	}
+
+	@Override
+	public void reset(int rlen, int[] odims, double v) {
```
mboehm7, line 69: It would be nice if we could reduce the code size of this method.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLBool.java
```diff
@@ -0,0 +1,366 @@
+/*
+ * Modifications Copyright 2018 Graz University of Technology
```
mboehm7: Generally please always use the year of the first modification.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLBool.java
```diff
@@ -0,0 +1,366 @@
...
+	@Override
+	public long capacity() {
+		return (_blocks!=null) ? (long)(_blocks.length - 1) * _blocks[0].size() + _blocks[_blocks.length - 1].size() : -1;
+	}
+
+	@Override
+	public long countNonZeros() {
+		long nnz = 0;
+		for (int i = 0; i < _blocks.length - 1; i++) {
+			nnz += UtilFunctions.computeNnz(_blocks[i], 0, blockSize() * _odims[0]);
```
mboehm7: if we ensure that all bits past the logical size of a block are zero, we could simply call Bitset.cardinality.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLBool.java
```diff
@@ -0,0 +1,366 @@
...
+		int destPos = 0;
+		int destBi = 0;
+		int finishedRows = 0;
+		// Loop through both blocks (source and destination) and always take the maximum number of rows that were not
+		// yet read from source or not yet written to destination. This procedure is necessary because although the dimensions
+		// should match there is the option that one DenseBlock was reset() and reused old blocks, therefore it is possible
+		// that one DenseBlock uses a block to represent x rows the other one can actually fit x+y rows in a block.
+		while (true) {
+			int length;
+			if (srcPos != 0) {
```
mboehm7: Is this really necessary - except for the boolean DenseBlock all of these dense blocks should be aligned as we systematically compute the blocksize. So maybe its a good idea to use the same blocksize for boolean blocks as well because several binary operations would have the same problem.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLBool.java
```diff
@@ -0,0 +1,366 @@
...
+				destPos = 0;
+				destBi++;
+			}
+		}
+		return this;
+	}
+
+	@Override
+	public DenseBlock set(int rl, int ru, int cl, int cu, DenseBlock db) {
+		boolean allColumns = cl == 0 && cu == _odims[0];
```
mboehm7: This allColumns variable is not used at all.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLFP32.java
```diff
@@ -0,0 +1,331 @@
...
+	public boolean isNumeric() {
+		return true;
+	}
+
+	@Override
+	public boolean isContiguous() {
+		return _blocks.length == 1;
+	}
+
+	private boolean isReusable(int rlen, int[] odims) {
```
mboehm7: While thinking about this notion of reusable, I remembered why before we only checked for exact matches: if we reuse too large blocks, the memory requirements of operations change and we might run out of memory although we explicitly scheduled an operation in memory. Accordingly, let's start with a more conservative approach to reusing. This might also make it much simpler such that we can pull it into the DenseBlock implementation.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockLInt32.java
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	@Override
+	public void reset(int rlen, int[] odims, double v) {
+		int iv = UtilFunctions.toInt(v);
+		if(!isReusable(rlen, odims)) {
+			// More memory is needed
+		    int newBlockSize = Integer.MAX_VALUE / odims[0];
+		    int restBlockSize = rlen % newBlockSize;
+		    int newNumBlocks = (rlen / newBlockSize) + (restBlockSize == 0 ? 0 : 1);
```
mboehm7: Let's think about what parts of these implementations are almost identical can can be moved to the base classes.

src/test/java/org/tugraz/sysds/test/component/tensor/DenseBlockSetRowTest.java
```diff
@@ -0,0 +1,262 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tugraz.sysds.test.component.tensor;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.tugraz.sysds.common.Types.ValueType;
+import org.tugraz.sysds.runtime.data.*;
```
mboehm7: Please avoid wildcard imports as it could make it hard to find the exact classes (especially if there are multiple wildcards and classes with the same name in different packages)

src/main/java/org/tugraz/sysds/runtime/util/UtilFunctions.java
```diff
@@ -621,7 +614,14 @@ public static int computeNnz(BitSet a, int ai, int len) {
 			lnnz += a.get(i) ? 1 : 0;
 		return lnnz;
 	}
-	
+
+	public static int computeNnz(String[] a, int ai, int len) {
+		int lnnz = 0;
+		for( int k=ai; k<ai+len; k++ )
+			lnnz += (a[k] != null && !a[k].isEmpty()) ? 1 : 0;
```
mboehm7: please also treat all Strings of value zero (0, 0.0, -0, +0, etc) as zero value.

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
```diff
@@ -51,6 +51,8 @@
 import org.tugraz.sysds.runtime.matrix.data.OutputInfo;
 import org.tugraz.sysds.runtime.meta.MatrixCharacteristics;
 
+import javax.rmi.CORBA.Util;
```
mboehm7: Unused import.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockFP64.java
```diff
@@ -187,4 +193,9 @@ public double get(int r, int c) {
 	public double get(int[] ix) {
 		return _data[pos(ix)];
 	}
+
+	@Override
+	public String getString(int[] ix) {
```
mboehm7, line 180: Apart from the String modifications, I guess we need to slightly generalize this implementation because otherwise it would not work for combinations of different value types (e.g., on set dense block). However, please be cautious as this double dense block is already in active use, so we need to make sure there are no performance regressions.

mboehm7 (commented):
Thanks @kev-inn for your hard work on this patch, which is a lot given the missing template support. Please address the comments, try to simplify the logic of reusing/allocation, and try to create abstractions for duplicated code wherever possible. Performance improvements can be postponed to another PR but please add related `TODO perf` comments.

mboehm7:
In addition to the inline comments, please make a pass over the code and remove warnings where possible (my IDE shows 48 additional warnings)

docs/Tasks.txt
kev-inn: I will watch out for that in the future.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockFP32.java
kev-inn: I added the Todos for performance, I guess this resolves this for now. Same for the DenseBlockBool

mboehm7:
hmm, "to reduce code reuse" ;-) let me know once you're happy with the PR and I'll have another look

kev-inn:
> hmm, "to reduce code reuse" ;-) let me know once you're happy with the PR and I'll have another look

my bad, of course it should say "to improve code reuse". The next commit should be the last one.

kev-inn:
@mboehm7 I think all issues you addressed should be resolved now. I have changed quite a bit to make the code easier to work with. Performance could be impacted, but not by much, if we consider the performance mistakes (iterating twice over array for copying, which was not necessary) it should actually have improved.

mboehm7:
LGTM - thanks @kev-inn for the patch. I only made some minor cleanups: formatting (tabs vs spaces, ternary operators, TODO instead of ToDo), replaced the used `NotImplementedException` with its counterpart in `apache.commons.lang`, and added some additional TODOs.

PR #17, opened 2019-07-18
Tensor Constructor, Sum and toString built-ins for DML

kev-inn:
This commit adds the built-in `tensor()` and support for tensors for the
built-ins `sum()` and `toString()`. The implementation is not very clean
and there are a lot of TODOs (which should all be marked by comments).

Note that the CostEstimatorStaticRuntime will probably not be possible to
accurately work on `tensor()` if we want to support all functionality of
`tensor()` currently implemented.

The `sum()` implementation is also not very in-depth and only very basic.

`toString()` works nicely, although we might have to think about how we
want to truncate the output if the size is too large. At the moment it
checks for the first two dimensions, but this of course does not really
limit the output size.

I was not sure at which point I should switch from a modifications license
to normal license by TU Graz, because I sometimes copied parts of already
implemented features.


kev-inn:
Note that at the moment only `Double` Tensors are supported (See `builtins-reference.md`). Are there any suggestions how I should support different types? Should there be another argument or should I infer the type from the data-argument?

mboehm7:
that is so awesome - thanks @kev-inn. I'll have a detailed look into this PR tomorrow evening.

src/main/java/org/tugraz/sysds/hops/TensorAggUnaryOp.java
```diff
@@ -0,0 +1,239 @@
...
+
+import org.tugraz.sysds.common.Types.DataType;
+import org.tugraz.sysds.common.Types.ValueType;
+import org.tugraz.sysds.lops.Lop;
+import org.tugraz.sysds.lops.LopProperties.ExecType;
+import org.tugraz.sysds.lops.TensorPartialAggregate;
+import org.tugraz.sysds.runtime.meta.MatrixCharacteristics;
+
+
+public class TensorAggUnaryOp extends MultiThreadedHop
```
mboehm7: I understand that this simplified the integration but I think we should reuse the existing hops and lops as otherwise there will be lots of redundancy. You can ignore the size propagation for know which we will generalize from 2D to nD once for all operators. By forcing the execution type to singlenode you can still experiment with your CP operations.

src/main/java/org/tugraz/sysds/hops/cost/CostEstimatorStaticRuntime.java
```diff
@@ -1,4 +1,6 @@
 /*
+ * Modifications Copyright 2019 Graz University of Technology
```
mboehm7, line 2: Whenever in doubt, please use this modifications license to properly acknowledge the original ASF sources.

src/main/java/org/tugraz/sysds/hops/cost/CostEstimatorStaticRuntime.java
```diff
@@ -424,7 +426,13 @@ else if( optype.equals("solve") ) //see also MultiReturnBuiltin
 						else
 							return xbu * d1m * d1n;
 					}
-										
+
+				case TensorGen: //opcodes: tensor
+					// TODO Cost estimation tensor
+					return 0;
+				case TensorAggregateUnary: //opcodes: sum with tensor as parameter
```
mboehm7: As stated above, please disregard all cost models, size propagation, and memory estimation components.

src/main/java/org/tugraz/sysds/hops/TensorGenOp.java
```diff
@@ -0,0 +1,200 @@
...
+		//mark for recompile (forever)
+		setRequiresRecompileIfNecessary();
+		
+		return _etype;
+	}
+	
+	@Override
+	public void refreshSizeInformation() {
+		Hop input1 = getInput().get(0);
+		Hop input2 = getInput().get(1); // dims
```
mboehm7: Commenting code and putting TODOs on them is fine, but please try to avoid warnings for unused code/variables, unnecessary imports, etc. You can simply comment the respective code, or use annotations to disable these warnings (but only if there is a good reason).

src/main/java/org/tugraz/sysds/lops/TensorPartialAggregate.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+package org.tugraz.sysds.lops;
+
+import org.tugraz.sysds.common.Types.DataType;
+import org.tugraz.sysds.common.Types.ValueType;
+import org.tugraz.sysds.hops.AggBinaryOp.SparkAggType;
+import org.tugraz.sysds.lops.LopProperties.ExecType;
+
+public class TensorPartialAggregate extends Lop
```
mboehm7: Just to elaborate a little more, I like to get rid of tensor-specific hops and lops because (1) many operators already work with scalars/matrices (so having separate ones for tensors would be inconsistent), and (2) for binary and nary operations we anyway have to deal with all combinations of input data types. Please try to branch out (create different in-meomry runtime instructions) during instruction parsing but not during instruction generation as they anyway have the input and output types compiled in them.

src/main/java/org/tugraz/sysds/parser/BuiltinFunctionExpression.java
```diff
@@ -568,7 +570,12 @@ public void validateExpression(HashMap<String, DataIdentifier> ids, HashMap<Stri
 		case VAR:
 			// sum(X);
 			checkNumParameters(1);
-			checkMatrixParam(getFirstExpr());
+			// TODO rewrite this
+			if (getOpCode() == Builtins.SUM) {
+				checkMatrixTensorParam(getFirstExpr());
```
mboehm7: Maybe it would be a good idea to annotate in Builtins which operations are supported over tensors, call checkMatrixTensorParam for all invocations, but check in that implementation if a particular builtin is indeed supported over tensors.

src/main/java/org/tugraz/sysds/parser/BuiltinFunctionExpression.java
```diff
@@ -1715,6 +1722,14 @@ protected void checkMatrixParam(Expression e) {
 		}
 	}
 
+	protected void checkMatrixTensorParam(Expression e) {
+		if (e.getOutput().getDataType() != DataType.MATRIX && e.getOutput().getDataType() != DataType.TENSOR) {
+			raiseValidateError(
+					"Expected " + e.getText() + " to be a matrix or tensor argument for function " + this.getOpCode().toString().toLowerCase() + "().",
+					false);
```
mboehm7: if you keep this, please reformat to readable form.

src/main/java/org/tugraz/sysds/parser/DMLTranslator.java
```diff
@@ -2255,6 +2263,11 @@ private Hop processBuiltinFunctionExpression(BuiltinFunctionExpression source, D
 			break;
 		
 		case SUM:
+			if (expr.getDataType() == DataType.TENSOR) {
```
mboehm7: if we reuse the hops and lops, this code in DMLTranslator could stay generic.

src/main/java/org/tugraz/sysds/parser/DataExpression.java
```diff
@@ -372,6 +439,40 @@ public void addMatrixExprParam(String paramName, Expression paramValue)
...
+				if (name.equals(paramName)) {
+					found = true;
+				}
+			}
+		}
+
+		if (!found){
+			raiseValidateError("unexpected parameter \"" + paramName +
+					"\". Legal parameters for tensor statement are "
+					+ "(capitalization-sensitive): " 	+ RAND_DATA + ", " + RAND_DIMS
```
mboehm7: avoid tabs in the middle of code, only for indentation

src/main/java/org/tugraz/sysds/parser/DataExpression.java
```diff
@@ -372,6 +439,40 @@ public void addMatrixExprParam(String paramName, Expression paramValue)
 		addVarParam(paramName,paramValue);
 	}
 
+	public void addTensorExprParam(String paramName, Expression paramValue)
+	{
+		// check name is valid
+		boolean found = false;
+		if (paramName != null ){
+			for (String name : TENSOR_VALID_PARAM_NAMES){
```
mboehm7: could we please rewrite such old code to java streams (i.e., anymatch)

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/ExecutionContext.java
```diff
@@ -203,6 +207,19 @@ public MatrixObject getMatrixObject(String varname) {
 		return (MatrixObject) dat;
 	}
 
+	public TensorBlockData getTensorObject(String varname) {
```
mboehm7, line 210: This is fine - and I will handle the buffer pool integration later this week

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -290,25 +293,93 @@ public double get(int[] ix) {
...
+			throw new NotImplementedException();
+		} else {
+			double sum = 0;
+			int elementsPerRow = Arrays.stream(_dims, 1, _dims.length).reduce(1, (a, b) -> a * b);
+			for (int bix = 0; bix < _denseBlock.numBlocks(); bix++) {
+				double[] values = _denseBlock.valuesAt(bix);
+				for (int r = 0; r < _denseBlock.blockSize(bix); r++) {
+					int finalR = r;
+					sum += IntStream.range(0, elementsPerRow).mapToDouble(i -> values[finalR * elementsPerRow + i]).sum();
+				}
```
mboehm7, line 401: As we want to use this prototype for benchmarking, make a TODO perf to remind us that we need to replace the use of streams here and make it multi-threaded as otherwise we'll not be able to saturate peak memory bandwidth.

src/main/java/org/tugraz/sysds/runtime/instructions/CPInstructionParser.java
```diff
@@ -362,7 +370,13 @@ public static CPInstruction parseSingleInstruction ( CPType cptype, String str )
 				
 			case Rand:
 				return DataGenCPInstruction.parseInstruction(str);
-				
+
+			case TensorGen:
+				return TensorGenCPInstruction.parseInstruction(str);
+
+			case TensorAggregateUnary:
```
mboehm7: given the lack of an opcode, you can check on AggregateUnary the input data type and then either call the respective matrix or tensor operations.

src/main/java/org/tugraz/sysds/runtime/instructions/cp/TensorGenCPInstruction.java
```diff
@@ -0,0 +1,179 @@
...
+		else
+			return new TensorGenCPInstruction(new Operator(true), in1, in2, in3, out, opcode, str);
+	}
+
+	@Override
+	public void processInstruction(ExecutionContext ec) {
+		TensorBlock out;
+		// TODO use byRow, or remove it
+		BooleanObject byRow = (BooleanObject) ec.getScalarInput(_opByRow.getName(), Types.ValueType.BOOLEAN,
+							_opByRow.isLiteral());
```
mboehm7: Please don't spend too much time on this random data generator. Instead please call the matrix random data generator and reshape it to a tensor for now. The reason is that we need to implement this data generator with awareness of the blocking of distributed tensors in order to ensure that rand with specified seed yields exactly the same result in both local and distributed operations (with blocked output representations).

src/main/java/org/tugraz/sysds/runtime/instructions/cp/VariableCPInstruction.java
```diff
@@ -523,6 +527,30 @@ else if ( getInput1().getDataType() == DataType.SCALAR ){
...
+				//(existing objects gets cleared through rmvar instructions)
+				//String fname = getInput2().getName();
+				// check if unique filename needs to be generated
+				//if( Boolean.parseBoolean(getInput3().getName()) ) {
+				//	fname = new StringBuilder(fname.length()+16).append(fname)
+				//			.append('_').append(_uniqueVarID.getNextID()).toString();
+				//}
+				// TODO Cacheable tensor block
+				// TODO TensorBlock constructor call with different ValueType (default FP64)
+				TensorBlockData tensor = new TensorBlockData(ValueType.FP64);
```
mboehm7: Please create it according to the value type in the instruction.

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
```diff
@@ -1,4 +1,6 @@
 /*
+ * Copyright 2019 Graz University of Technology
```
mboehm7: Modifications license.

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
```diff
@@ -904,6 +908,149 @@ public static String toString(MatrixBlock mb, boolean sparse, String separator,
...
+					for (int item : ix) {
+						sb.append(item).append(separator);
+					}
+					concatenateTensorValue(tb, sb, df, ix);
+					sb.append(lineseparator);
+				}
+				ix[j]++;
+				//calculating next index
+				if (ix[j] == tb.getDim(j)) {
+					while (ix[j] == tb.getDim(j) || ix[1] >= colLength) {
```
mboehm7: I've seen a similar code fragment before, so please either (1) simplify to some easier-to-understand for loop, or (2) encapsulate it in a function for index computation.

mboehm7 (commented):
As I said before, this is an awesome start @kev-inn. Let's try to collapse the replicated hops/lop code and get a basic version of this patch into master to start with the performance experiments and improvements.

src/main/java/org/tugraz/sysds/parser/BuiltinFunctionExpression.java
kev-inn: I copied this from `checkMatrixParam` and adapted it, should I also reformat other methods that also? (In general I tried to stay as close as possible to already implemented methods, should I adapt them if I think it would lead to more readable code?)

src/main/java/org/tugraz/sysds/parser/DataExpression.java
kev-inn: Should this instead be longer than 120 characters?

src/main/java/org/tugraz/sysds/runtime/instructions/cp/TensorGenCPInstruction.java
kev-inn: I am actually not using a random data generator, I think I just misnamed the instruction. `TensorGen` is responsible to construct the tensor from some kind of given data, which could be a string, matrix, tensor or single scalar. We can discuss a new name in today's meeting.

src/main/java/org/tugraz/sysds/lops/TensorPartialAggregate.java
kev-inn: I have removed it, but concerning branching during instruction parsing: should I create a different CPInstruction object or should I just implement the calculation inside `AggregateUnaryCPInstruction`, I did the latter for now.

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
kev-inn: I encapsulated the loop for the sparse print format, for dense though using the encapsulated method leads to even harder to understand code so I simplified it a bit and added comments. I hope it is now clear enough.

kev-inn:
By the way this should complete this PR. If there are new issues, which is likely since a lot got rewritten, just tell me and I will try to fix it.

kev-inn:
I have rewritten the sum to use multi-threaded kahan sum, the way matrix use it. I tried to keep my implementation similar to the already implemented matrix aggregation, with simplifications, because only sum is supported and sparse is not yet completely implemented for tensor. Should I add this addition to this PR or wait until this one is merged?
Edit: I might add that the kahan sum currently is only intended to work by reducing the tensor to a single value

mboehm7:
I intend to merge this PR here today, so it would be awesome if you could put performance-related improvements into another PR. Thanks.

kev-inn:
Okay, I will make a separate PR and maybe add some other performance improvements.

mboehm7:
LGTM - thanks a lot @kev-inn. Everything was fine - I just made a few syntactic changes (fixed warnings, refresh size info, and removed unused tensor types). 

Just for the future: please try to use refactorings + commit for renamings like `MatrixReshapeCPInstruction` to `ReshapeCPInstruction` because otherwise we loose the history if git does not recognize that it was a rename plus modifications.

kev-inn:
Thanks for the tip, I did not notice that it was not recognizing the renaming.

PR #18, opened 2019-07-26
Add multi-threaded kahan sum aggregation. Optimize aggregation for different datatypes

kev-inn:
This commit improves the sum operation to use multiple threads. This is as fast as the current matrix implementation for all datatypes except String, which is slower and it should be discussed how a `sum` on Strings should even work, and Boolean, where it is a lot faster. If the matrix is empty though, it optimizes it to be a metadata operation (with non zero count), while the tensor operation does not yet support this.

Reshape instruction should also be as fast as matrix reshape and makes use of copy on write.

Note that all instructions are only implemented for CP-mode.


kev-inn:
Concerning the non optimized empty aggregation, I figured out the problem, the variable holding the tensor somehow is not put into the `LocalVariableMap`. I have yet to figure out why this is the case and where the code handling this for matrices is located.

mboehm7:
so you didn't try out INT32 and FP32 yet, right? For these data types, we should also be about 2x faster.

kev-inn:
No, I did not, since I do not know how I can instantiate a literal of that type in DML, I did not find anything in that respect in the SystemML documentation. Is there a way in SystemDS?
Therefore I did also not optimize for those cases so it could be slow.

mboehm7:
I would recommend, to just temporarily cast the double input tensor to the respective type inside your sum implementation and measure just the time for executing the sum (without the conversion).

src/main/java/org/tugraz/sysds/runtime/data/DataTensorBlock.java
```diff
@@ -1,23 +0,0 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package org.tugraz.sysds.runtime.data;
-
-public class DataTensorBlock extends TensorBlock
```
mboehm7: Please keep this as a reminder of implementing DataTensors, which can be composed of tensor blocks of certain types plus a column mapping. I'd like to early on decide if we should implement operations in TensorBlock and then composite operations in DataTensorBlock or if we should directly focus on DataTensorBlock.

src/main/java/org/tugraz/sysds/runtime/data/DenseBlockBool.java
```diff
@@ -44,12 +44,13 @@ protected void allocateBlock(int bix, int length) {
 		_data = new BitSet(length);
 	}
 
-	public DenseBlockBool(int[] dims, boolean[] data) {
```
mboehm7: Could we please keep both constructors, as passing a boolean[] array might be useful for conversion purposes.

src/main/java/org/tugraz/sysds/runtime/data/LibTensorAgg.java
```diff
@@ -0,0 +1,396 @@
...
+	 */
+	private static void aggregateBinaryTensorLastRowDenseGeneric(TensorBlock in, TensorBlock aggVal) {
+		if( in._denseBlock==null || in.isEmpty(false) )
+			return;
+
+		final int m = in.getDim(0);
+		final int n = in.getDim(1);
+		final int cix = (m-1)*n;
+
+		double[] a = in.getDenseBlock().valuesAt(0);
```
mboehm7: Please add a TODO for generalize for large blocks if you use it for local operations of large matrices (not vectors).

src/main/java/org/tugraz/sysds/runtime/data/LibTensorAgg.java
```diff
@@ -0,0 +1,396 @@
...
+		} else {
+			// TODO special case String: currently converted to double, should be concatenated?
+			// Less readable and makes use of how row-column-getter actually works, but is a lot faster than
+			// using the index[]-getter.
+			long sum = 0;
+			for (int bix = bil; bix <= biu; bix++) {
+				int lpos = (bix == bil) ? a.pos(rl) : 0;
+				int len = (bix == biu) ? a.pos(ru - 1) - lpos + a.getCumODims(0) : a.blockSize(bix) * a.getCumODims(0);
+				for (int i = 0; i < len; i++)
+					sum += UtilFunctions.toLong(a.get(bix * a.blockSize() * a.getCumODims(0), lpos + i));
```
mboehm7: Did you try to iterate in two loops, rows and other dimensions (from cumODims) and doing a simple get(i,j)? Also, why is there a toLong cast in the case that also applies to FP32?

mboehm7 (commented):
Overall, this is an excellent start @kev-inn - let's do some cleanup and then we can merge it in. Before we go further ahead with these operations though, we should have our next call to make sure we're all happy with the general template. It would be awesome to find a clean, understandable, general-case implementation (for all remaining data types) that closely matches the expected performance for memory-bandwidth bound operations.

src/main/java/org/tugraz/sysds/runtime/data/DataTensorBlock.java
kev-inn: Sorry, of course. I deleted this file on accident, because I mixed it up with my temporary Data extended TensorBlock

src/main/java/org/tugraz/sysds/runtime/data/LibTensorAgg.java
kev-inn: with the simple get(i,j) are you refering to the FP64 case?
My mistake, I should of course NOT cast toLong.

mboehm7: I was thinking of something like https://github.com/tugraz-isds/systemds/blob/master/src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java#L371 (except that we don't even need a switch if we simply compute a double sum) as a fallback for all data types (except string) where we do not want to integrate a custom implementation.

kev-inn:
I fixed the issues, but I did not yet do in depth performance tests for int32 or float. The tests I did showed that performance is similar to long and double, so there should be space for improvement. I will think about how we can make the current implementation a bit cleaner, while not sacrificing performance, so we have a nice template for future implementations.

mboehm7:
LGTM - thanks @kev-inn - I only made minor changes to remove unnecessary warnings and fix an issue in `TensorBlock.copy()`

PR #20, opened 2019-07-30
[SYSTEMDS-75] Minor fixes for lineage tracing and full-reuse approach

bnyra:
Hey @mboehm7,

I have some small changes for lineage tracing:

 * Fix CLI bug for `-lineage dedup` command
 * Add `LineageTracable` interface for `ComputationSPInstruction` 
 * Move `Lineage.trace()` from `postprocessInstruction()` into `preprocessInstruction()`
 * Adapt runtimeSeed generation for `DataGenCPInstruction`
 * Add simple full-reuse of already computed instructions by `LineageItem` as key

### DIFF
--- a/src/main/java/org/tugraz/sysds/api/DMLOptions.java
+++ b/src/main/java/org/tugraz/sysds/api/DMLOptions.java
@@ -106,7 +106,10 @@ public class DMLOptions {
 			dmlOptions.lineage = true;
 			String lineageType = line.getOptionValue("lineage");
 			if (lineageType != null){
-				dmlOptions.lineage_dedup = lineageType.equalsIgnoreCase("dedup");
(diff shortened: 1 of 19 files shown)


mboehm7:
LGTM - thanks @bnyra. I some minor modifications, including (1) a change of `CPInstruction.preprocessInstruction` to move the default preprocess (which now includes lineage tracing) after instruction patching, and (2) a missing reset of the runtime seed to ensure we're still generating random seeds for every invocation in loops.

mboehm7:
Also, we'll modify the cache implementation from Data to CacheBlock objects to ensure the matrices are kept around; otherwise the data would be cleaned up on the next rmvar instruction and then it fails (e.g., one of the lineage tests right now). Anyway this is a great start, thanks for your effort, and Arnab will take over from here.

PR #21, opened 2019-07-31
[SYSTEMDS-75] Adapt LineageCache to MatrixBlock

bnyra:
* Changed the value for `LineageCache` from `Data` to `MatrixBlock`
    Only enabled for ComputationCPInstruction which returns a `MatrixObject` as output
 * Add cli parameter `-lineage reuse' for check soundness of lineage-based reusing
 * Add additional `CPOperand` tracing for `MatrixIndexingCPInstruction`


mboehm7:
LGTM - thanks @bnyra; I only made a small syntactic change in `MatrixIndexingCPInstruction.getLineageItems`

PR #22, opened 2019-08-02
Stepwise linear regression builtin function

iulian453:
New commit with changes, apparently the function stop() was stopping all the process (without returning anything back), so I just changed it with an "else".


mboehm7:
LGTM - thanks @iulian453 for the fix and cleanups. This works like a charm now. I made some minor changes though: (1) removed forced singlenode runtime in tests, and (2) removed non-englisch prints in R script. Also in the future, please create PRs only over the new commit not over all commits (this PR included everything from PR 19).

PR #23, opened 2019-08-05
Initial version of lineage cache with statistics.

phaniarnab:
Find the performance data below for 10000*1024 matrix. Result is similar for both forward and backward stepwise loops.

**1. Forward loop**
- **With lineage and without reuse**
SystemML Statistics:
Total elapsed time:		        631.563 sec.
Total execution time:		629.951 sec.
  1  tsmm         477.466   4092
 

- **With lineage and with reuse**
SystemML Statistics:
Total elapsed time:		                 300.522 sec.
Total execution time:		         298.888 sec.
LineageCache hits (Mem/FS/Del):   2044/642/382.
LineageCache writes (Mem/FS): 	2048/1216.
LineageCache FStimes (Rd/Wr): 	9.169/12.696 sec.
LineageCache costing time: 	        0.007 sec.
  1  tsmm         136.456   4092

**2. Backward loop**
- **With lineage and without reuse**
SystemML Statistics:
Total elapsed time:		616.341 sec.
Total execution time:		614.790 sec.
  1  tsmm         450.241   4092

- **With lineage and with reuse**
SystemML Statistics:
Total elapsed time:		293.870 sec.
Total execution time:		292.334 sec.
LineageCache hits (Mem/FS/Del): 2044/642/382.
LineageCache writes (Mem/FS): 	2048/1081.
LineageCache FStimes (Rd/Wr): 	8.591/11.420 sec.
LineageCache costing time: 	0.010 sec.
  1  tsmm         123.887   4092

Note: In LineageCache hits, 'Del' specifies number of cache misses which should have been found in cache but removed earlier due to lack of space.

### DIFF
--- a/src/main/java/org/tugraz/sysds/runtime/instructions/Instruction.java
+++ b/src/main/java/org/tugraz/sysds/runtime/instructions/Instruction.java
@@ -25,7 +25,14 @@ import org.tugraz.sysds.api.DMLScript;
 import org.tugraz.sysds.lops.Lop;
 import org.tugraz.sysds.parser.DataIdentifier;
 import org.tugraz.sysds.runtime.controlprogram.context.ExecutionContext;
+import org.tugraz.sysds.runtime.instructions.cp.CPOperand;
+import org.tugraz.sysds.runtime.instructions.cp.ComputationCPInstruction;
+import org.tugraz.sysds.runtime.instructions.cp.Data;
+import org.tugraz.sysds.runtime.instructions.spark.ComputationSPInstruction;
 import org.tugraz.sysds.runtime.lineage.Lineage;
+import org.tugraz.sysds.runtime.lineage.LineageCache;
+import org.tugraz.sysds.runtime.lineage.LineageItem;
+import org.tugraz.sysds.runtime.lineage.LineageTraceable;
 
 
 public abstract class Instruction 
@@ -223,7 +230,6 @@ public abstract class Instruction
 		//return instruction ifself
 		return this;
 	}
-	
 	/**
 	 * This method should be used to execute the instruction. 
 	 * 
--- a/src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
+++ b/src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
@@ -1,60 +1,307 @@
+/*
+ * Copyright 2019 Graz University of Technology
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.tugraz.sysds.runtime.lineage;
 
 import org.tugraz.sysds.api.DMLScript;
+import org.tugraz.sysds.hops.OptimizerUtils;
+import org.tugraz.sysds.hops.cost.CostEstimatorStaticRuntime;
+import org.tugraz.sysds.lops.MMTSJ.MMTSJType;
+import org.tugraz.sysds.runtime.DMLRuntimeException;
 import org.tugraz.sysds.runtime.controlprogram.caching.MatrixObject;
 import org.tugraz.sysds.runtime.controlprogram.context.ExecutionContext;
+import org.tugraz.sysds.runtime.instructions.CPInstructionParser;
 import org.tugraz.sysds.runtime.instructions.Instruction;
+import org.tugraz.sysds.runtime.instructions.cp.CPInstruction.CPType;
 import org.tugraz.sysds.runtime.instructions.cp.ComputationCPInstruction;
+import org.tugraz.sysds.runtime.instructions.cp.MMTSJCPInstruction;
 import org.tugraz.sysds.runtime.matrix.data.MatrixBlock;
+import org.tugraz.sysds.runtime.util.LocalFileUtils;
 
+import java.io.IOException;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 public class LineageCache {
-	private static final Map<LineageItem, MatrixBlock> _cache = new HashMap<>();
-	
+	private static final Map<LineageItem, Entry> _cache = new HashMap<>();
+	private static final Map<LineageItem, String> _spillList = new HashMap<>();
+	private static final HashSet<LineageItem> _removelist = new HashSet<>();
+	private static final long CACHELIMIT= (long)512*1024*1024; // 500MB
+	private static String outdir = null;
+	private static long _cachesize = 0;
+	private static Entry _head = null;
+	private static Entry _end = null;
+
+	//--------------------- CACHE LOGIC METHODS ----------------------
+
 	public static void put(Instruction inst, ExecutionContext ec) {
 		if (!DMLScript.LINEAGE_REUSE)
 			return;
-		
-		if( inst instanceof ComputationCPInstruction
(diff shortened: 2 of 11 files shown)


src/test/java/org/tugraz/sysds/test/functions/lineage/LineageRewriteTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.junit.Test;
+import org.tugraz.sysds.runtime.controlprogram.caching.MatrixObject;
+import org.tugraz.sysds.runtime.instructions.cp.Data;
+import org.tugraz.sysds.runtime.instructions.cp.ScalarObject;
+import org.tugraz.sysds.runtime.lineage.LineageItem;
+import org.tugraz.sysds.runtime.lineage.LineageItemUtils;
+import org.tugraz.sysds.runtime.lineage.LineageParser;
+import org.tugraz.sysds.runtime.matrix.data.MatrixBlock;
+import org.tugraz.sysds.runtime.matrix.data.MatrixValue.CellIndex;
```
mboehm7, line 31: Please remove unnecessary imports

src/main/java/org/tugraz/sysds/runtime/instructions/Instruction.java
```diff
@@ -25,7 +25,14 @@
 import org.tugraz.sysds.lops.Lop;
 import org.tugraz.sysds.parser.DataIdentifier;
 import org.tugraz.sysds.runtime.controlprogram.context.ExecutionContext;
+import org.tugraz.sysds.runtime.instructions.cp.CPOperand;
+import org.tugraz.sysds.runtime.instructions.cp.ComputationCPInstruction;
+import org.tugraz.sysds.runtime.instructions.cp.Data;
+import org.tugraz.sysds.runtime.instructions.spark.ComputationSPInstruction;
```
mboehm7, line 31: Please remove unnecessary imports.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
+	private static final HashSet<LineageItem> _removelist = new HashSet<>();
+	private static final long CACHELIMIT= (long)512*1024*1024; // 500MB
+	private static String outdir = null;
+	private static long _cachesize = 0;
+	private static Entry _head = null;
+	private static Entry _end = null;
+
+	//--------------------- CACHE LOGIC METHODS ----------------------
+
 	public static void put(Instruction inst, ExecutionContext ec) {
```
mboehm7, line 51: We should synchronize this put/get - for now you can synchronize the entire method, once we optimize performance, we'll likely move it around the critical data structure updates only to reduce the critical section.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
 		if (!DMLScript.LINEAGE_REUSE)
 			return;
-		
-		if( inst instanceof ComputationCPInstruction
-			&&((ComputationCPInstruction) inst).output.getDataType().isMatrix() ) {
-			
-			for (LineageItem item : ((LineageTraceable) inst).getLineageItems()) {
+
+		if (inst instanceof ComputationCPInstruction &&
+			inst.getOpcode().equalsIgnoreCase("tsmm")) {
```
mboehm7, line 56: Hard-coding tsmm here is ok for now, but please already move this check to a dedicated function or class where we will implement various configurations such as 'full reuse', 'reuse only annotated', 'reuse only certain opcodes', etc

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
+			LineageItem[] items = ((LineageTraceable) inst).getLineageItems();
+			for (LineageItem item : items) {
 				MatrixObject mo = ec.getMatrixObject(((ComputationCPInstruction) inst).output);
-				LineageCache._cache.put(item, mo.acquireReadAndRelease());
+				LineageCache.put(inst, item, mo.acquireRead());
```
mboehm7, line 60: Please use the shorter `mo.acquireReadAndRelease()`

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
 
+import java.io.IOException;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 public class LineageCache {
-	private static final Map<LineageItem, MatrixBlock> _cache = new HashMap<>();
-	
+	private static final Map<LineageItem, Entry> _cache = new HashMap<>();
```
mboehm7, line 40: Consider using a `LinkedHashMap` here to avoid the custom Entry implementation. I guess you could emulate LRU by doing another put which should place the object at the end of the queue.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
mboehm7: Also please add a comment that it's safe to pin the object here, because it comes from a CPInstruction and hence is guaranteed to fit into the driver.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
+			// Read and put the entry at head.
+			Entry e = _cache.get(key);
+			delete(e);
+			setHead(e);
+			if (DMLScript.STATISTICS)
+				LineageCacheStatistics.incrementMemHits();
+			return e._value;
+		}
+		else
+			return readFromLocalFS(inst, key);
```
mboehm7, line 142: Doesn't this call fail if the key was never evicted? Unless I'm missing something you might need to probe the spill list first.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
```diff
@@ -1,60 +1,307 @@
...
+	private static void removeEntry(double space) {
+		if (DMLScript.STATISTICS)
+			_removelist.add(_end._key);
+		_cache.remove(_end._key);
+		_cachesize -= space;
+		delete(_end);
+	}
+	
+	private static class Entry {
+		LineageItem _key;
```
mboehm7, line 297: It would be great if we could remove this class as there are way too many `Entry` implementations already.

src/main/java/org/tugraz/sysds/runtime/util/LocalFileUtils.java
```diff
@@ -65,6 +65,7 @@
 	public static final String CATEGORY_RESULTMERGE  = "resultmerge";
 	public static final String CATEGORY_WORK         = "work";
 	public static final String CATEGORY_CODEGEN      = "codegen";
+	public static final String CATEGORY_LINEAGE		 = "lineage";
```
mboehm7, line 68: The formatting is off - please use spaces inside code lines (but tabs at the beginning)

src/test/scripts/functions/lineage/RewriteTest1.dml
```diff
@@ -0,0 +1,24 @@
+X = read($1);
```
mboehm7, line 1: The license header is missing.

mboehm7 (commented):
Thanks @phaniarnab - this is awesome. There are only some minor cleanups needed.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
phaniarnab: As I realized, `LinkedHashMap` is a cleaner replacement of the custom data structure, but I have a feeling that I might need to use the LinkedList nodes to store few more things.  If that doesn't turn out to be the case, I will replace in next couple of PRs.

src/main/java/org/tugraz/sysds/runtime/lineage/LineageCache.java
phaniarnab: As discussed, execution will reach here only when spilllist contains this item.

phaniarnab:
Created a new pull request with the fixes.

PR #24, opened 2019-08-05
Add setter and getter for long values to TensorBlock

kev-inn:
This adds getter and setter for Long values to TensorBlock and DenseBlocks.


mboehm7:
LGTM - thanks for the patch @kev-inn. I only fixed the `TensorBlock.isEmpty()` to properly check for the nnz because otherwise we forget about it once this meta data is maintained.

PR #25, opened 2019-08-05
Add spark rand and full sum

kev-inn:
Note that there is still a bug for `TensorConstructionTest` case [6], the string will be shown to contain "null" while it should contain "test". I will fix this ASAP, but while I am working on it you can take a look at the other things.

### DIFF
--- a/src/main/java/org/tugraz/sysds/api/mlcontext/FrameMetadata.java
+++ b/src/main/java/org/tugraz/sysds/api/mlcontext/FrameMetadata.java
@@ -474,10 +474,10 @@ public class FrameMetadata extends Metadata {
 
 	/**
 	 * Constructor to create a FrameMetadata object based on a
-	 * SpaceCharacteristics object.
+	 * DataCharacteristics object.
 	 *
 	 * @param dataCharacteristics
-	 *            the frame metadata as a SpaceCharacteristics object
+	 *            the frame metadata as a DataCharacteristics object
 	 */
 	public FrameMetadata(DataCharacteristics dataCharacteristics) {
 		this.numRows = dataCharacteristics.getRows();
@@ -489,12 +489,12 @@ public class FrameMetadata extends Metadata {
 
 	/**
 	 * Constructor to create a FrameMetadata object based on the frame schema
-	 * and a SpaceCharacteristics object.
+	 * and a DataCharacteristics object.
 	 *
 	 * @param frameSchema
 	 *            The frame schema.
 	 * @param matrixCharacteristics
-	 *            the frame metadata as a SpaceCharacteristics object
+	 *            the frame metadata as a DataCharacteristics object
 	 */
 	public FrameMetadata(FrameSchema frameSchema, MatrixCharacteristics matrixCharacteristics) {
 		this.frameSchema = frameSchema;
@@ -506,10 +506,10 @@ public class FrameMetadata extends Metadata {
 	}
 
 	/**
-	 * Set the FrameMetadata fields based on a SpaceCharacteristics object.
+	 * Set the FrameMetadata fields based on a DataCharacteristics object.
 	 *
 	 * @param matrixCharacteristics
-	 *            the frame metadata as a SpaceCharacteristics object
+	 *            the frame metadata as a DataCharacteristics object
 	 */
 	@Override
 	public void setMatrixCharacteristics(MatrixCharacteristics matrixCharacteristics) {
--- a/src/main/java/org/tugraz/sysds/api/mlcontext/FrameMetadata.java
+++ b/src/main/java/org/tugraz/sysds/api/mlcontext/FrameMetadata.java
@@ -1,4 +1,6 @@
 /*
+ * Modifications Copyright 2019 Graz University of Technology
+ *
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -19,6 +21,7 @@
 
 package org.tugraz.sysds.api.mlcontext;
 
+import org.tugraz.sysds.runtime.meta.DataCharacteristics;
 import org.tugraz.sysds.runtime.meta.MatrixCharacteristics;
 
 /**
@@ -471,27 +474,27 @@ public class FrameMetadata extends Metadata {
 
 	/**
 	 * Constructor to create a FrameMetadata object based on a
-	 * MatrixCharacteristics object.
+	 * SpaceCharacteristics object.
 	 *
-	 * @param matrixCharacteristics
-	 *            the frame metadata as a MatrixCharacteristics object
+	 * @param dataCharacteristics
+	 *            the frame metadata as a SpaceCharacteristics object
 	 */
-	public FrameMetadata(MatrixCharacteristics matrixCharacteristics) {
-		this.numRows = matrixCharacteristics.getRows();
-		this.numColumns = matrixCharacteristics.getCols();
-		this.numNonZeros = matrixCharacteristics.getNonZeros();
-		this.numRowsPerBlock = matrixCharacteristics.getRowsPerBlock();
-		this.numColumnsPerBlock = matrixCharacteristics.getColsPerBlock();
+	public FrameMetadata(DataCharacteristics dataCharacteristics) {
+		this.numRows = dataCharacteristics.getRows();
+		this.numColumns = dataCharacteristics.getCols();
+		this.numNonZeros = dataCharacteristics.getNonZeros();
+		this.numRowsPerBlock = dataCharacteristics.getRowsPerBlock();
+		this.numColumnsPerBlock = dataCharacteristics.getColsPerBlock();
 	}
 
 	/**
 	 * Constructor to create a FrameMetadata object based on the frame schema
-	 * and a MatrixCharacteristics object.
+	 * and a SpaceCharacteristics object.
 	 *
 	 * @param frameSchema
 	 *            The frame schema.
 	 * @param matrixCharacteristics
-	 *            the frame metadata as a MatrixCharacteristics object
+	 *            the frame metadata as a SpaceCharacteristics object
 	 */
 	public FrameMetadata(FrameSchema frameSchema, MatrixCharacteristics matrixCharacteristics) {
 		this.frameSchema = frameSchema;
@@ -503,10 +506,10 @@ public class FrameMetadata extends Metadata {
 	}
 
 	/**
-	 * Set the FrameMetadata fields based on a MatrixCharacteristics object.
+	 * Set the FrameMetadata fields based on a SpaceCharacteristics object.
 	 *
 	 * @param matrixCharacteristics
-	 *            the frame metadata as a MatrixCharacteristics object
+	 *            the frame metadata as a SpaceCharacteristics object
 	 */
 	@Override
 	public void setMatrixCharacteristics(MatrixCharacteristics matrixCharacteristics) {
--- a/src/main/java/org/tugraz/sysds/parser/DataExpression.java
+++ b/src/main/java/org/tugraz/sysds/parser/DataExpression.java
@@ -678,6 +678,7 @@ public class DataExpression extends DataIdentifier
 			
 			// replace DataOp MATRIX with RAND -- Rand handles matrix generation for Scalar values
 			// replace data parameter with min / max within Rand case below
+			// TODO either use dims or rows, cols depending on datatype
 			BooleanIdentifier id = new BooleanIdentifier(_opcode == DataOp.TENSOR, this);
 			addVarParam(RAND_TENSOR, id);
 			this.setOpCode(DataOp.RAND);
@@ -1181,10 +1182,8 @@ public class DataExpression extends DataIdentifier
 				}
 				// handle integer constant 
 				else if (dataParam instanceof IntIdentifier) {
-					long roundedValue = ((IntIdentifier)dataParam).getValue();
-					Expression minExpr = new DoubleIdentifier(roundedValue, this);
-					addVarParam(RAND_MIN, minExpr);
-					addVarParam(RAND_MAX, minExpr);
+					addVarParam(RAND_MIN, dataParam);
+					addVarParam(RAND_MAX, dataParam);
 				}
 				// handle double constant 
 				else if (dataParam instanceof DoubleIdentifier) {
--- a/src/main/java/org/tugraz/sysds/parser/DataExpression.java
+++ b/src/main/java/org/tugraz/sysds/parser/DataExpression.java
@@ -21,13 +21,6 @@
 
 package org.tugraz.sysds.parser;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map.Entry;
-
 import org.antlr.v4.runtime.ParserRuleContext;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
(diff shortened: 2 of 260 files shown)


src/main/java/org/tugraz/sysds/api/mlcontext/FrameMetadata.java
```diff
@@ -471,27 +474,27 @@ public FrameMetadata(int numRows, int numColumns, int numNonZeros, int numRowsPe
 
 	/**
 	 * Constructor to create a FrameMetadata object based on a
-	 * MatrixCharacteristics object.
+	 * SpaceCharacteristics object.
```
mboehm7: Please conduct a search for SpaceCharacteristics and globally replace it with DataCharacteristics

src/main/java/org/tugraz/sysds/runtime/instructions/spark/RandSPInstruction.java
```diff
@@ -19,13 +21,7 @@
...
 package org.tugraz.sysds.runtime.instructions.spark;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.Random;
-
+import org.apache.commons.lang.NotImplementedException;
```
mboehm7: Remove unnecessary imports.

src/main/java/org/tugraz/sysds/runtime/instructions/cp/CPInstruction.java
```diff
@@ -23,27 +23,20 @@
 
 import org.tugraz.sysds.api.DMLScript;
 import org.tugraz.sysds.common.Types;
```
mboehm7: REmove unnecessary imports.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -418,6 +443,77 @@ else if(inputInfo == InputInfo.BinaryCellInputInfo) {
...
+				_parRDDs.registerRDD(rdd.id(), OptimizerUtils.estimatePartitionedSizeExactSparsity(dc), true);
+			}
+
+			//keep rdd handle for future operations on it
+			RDDObject rddhandle = new RDDObject(rdd);
+			rddhandle.setHDFSFile(fromFile);
+			rddhandle.setParallelizedRDD(!fromFile);
+			to.setRDDHandle(rddhandle);
+		}
+		//CASE 3: non-dirty (file exists on HDFS)
```
mboehm7, line 497: Fix the type safety warnings.

src/main/java/org/tugraz/sysds/parser/DataExpression.java
```diff
@@ -665,14 +671,15 @@ public void validateExpression(HashMap<String, DataIdentifier> ids, HashMap<Stri
...
-		Boolean tensorOperation = _opcode == DataOp.TENSOR;
 		if (dataParam1 != null && dataParam1.getOutput().getDataType() == DataType.SCALAR &&
 				(_opcode == DataOp.MATRIX || _opcode == DataOp.TENSOR)/*&& dataParam instanceof ConstIdentifier*/ ){
 			//MB: note we should not check for const identifiers here, because otherwise all matrix constructors with
 			//variable input are routed to a reshape operation (but it works only on matrices and hence, crashes)
 			
 			// replace DataOp MATRIX with RAND -- Rand handles matrix generation for Scalar values
 			// replace data parameter with min / max within Rand case below
+			BooleanIdentifier id = new BooleanIdentifier(_opcode == DataOp.TENSOR, this);
+			addVarParam(RAND_TENSOR, id);
```
mboehm7, line 683: It's ok for now, but later we simply allow either dims or rows/cols as arguments.

src/main/java/org/tugraz/sysds/runtime/controlprogram/caching/TensorObject.java
```diff
@@ -71,23 +77,31 @@ public void refreshMetaData() {
...
 		//update matrix characteristics
-		MatrixCharacteristics mc = _metaData.getMatrixCharacteristics();
-		mc.setDimension( _data.getNumRows(),_data.getNumColumns() );
-		mc.setNonZeros(_data.getNumRows()*_data.getNumColumns());
+		DataCharacteristics tc = _metaData.getDataCharacteristics();
+		long[] dims = new long[_data.getNumDims()];
+		for (int i = 0; i < _data.getNumDims(); i++) {
+			dims[i] = _data.getDim(i);
+		}
+		tc.setDims(dims);
```
mboehm7, line 85: Could you please implement a `set(DataCharacteristics)` instead?

src/main/java/org/tugraz/sysds/runtime/controlprogram/caching/TensorObject.java
```diff
@@ -107,8 +121,13 @@ protected TensorBlock readBlobFromHDFS(String fname, long rlen, long clen)
 	protected TensorBlock readBlobFromRDD(RDDObject rdd, MutableBoolean status)
 		throws IOException 
 	{
+		status.setValue(false);
+		TensorBlock tb;
```
mboehm7: Please avoid pure declarations whenever possible.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -732,8 +828,54 @@ public void setRDDHandleForVariable(String varname, JavaPairRDD<?,?> rdd) {
...
-	
-	private static Tuple2<MatrixIndexes,MatrixBlock> createIndexedBlock(MatrixBlock mb, MatrixCharacteristics mc, long ix) {
+
+	public static JavaPairRDD<TensorIndexes, TensorBlock> toTensorJavaPairRDD(JavaSparkContext sc, TensorBlock src,
+	                                                                          int[] blen) {
+		return toTensorJavaPairRDD(sc, src, blen, -1, true);
+	}
+
+	public static JavaPairRDD<TensorIndexes, TensorBlock> toTensorJavaPairRDD(JavaSparkContext sc, TensorBlock src,
+	                                                                          int[] blen, int numParts,
```
mboehm7: please bring the parameters on one line with two tabs indentation.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -755,6 +897,33 @@ public void setRDDHandleForVariable(String varname, JavaPairRDD<?,?> rdd) {
...
+			for (int i = tc.getNumDims() - 1; i >= 0; i--) {
+				blockIx[i] = ix % tc.getNumBlocks(i);
+				maxDim[i] = UtilFunctions.computeBlockSize(tc.getDim(i), blockIx[i] + 1,
+						tc.getBlockSize(i));
+				offset[i] = (int) (blockIx[i] * tc.getBlockSize(i));
+				ix /= tc.getNumBlocks(i);
+			}
+			// TODO: sparse
+			TensorBlock block = new TensorBlock(mb.getValueType(), maxDim, false);
+			block = mb.slice(offset, block);
```
mboehm7: How are the offsets enough to do proper blocking when all TensorBlocks are create with the same maxDim - specifically does it correctly handle boundary blocks of different sizes?

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -525,6 +545,23 @@ private void copyDenseToDense(TensorBlock that) {
 		_denseBlock.set(that._denseBlock);
 	}
 
+	public void copy(long[] offsets, TensorBlock src) {
```
mboehm7: consider using int[] lower and int[] upper as arguments as otherwise it will be hard to track bugs where the src does not match the indexing range.

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -636,9 +673,46 @@ public CacheBlock slice(int rl, int ru, int cl, int cu, CacheBlock block) {
 		return null;
 	}
 
+	public TensorBlock slice(int[] offsets, TensorBlock block) {
+		assert offsets.length == block.getNumDims();
```
mboehm7: please remove all asserts as they are cluttering the code and are anyway only checked if -ea is passed to the JVM (which it never is in practice)

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -636,9 +673,46 @@ public CacheBlock slice(int rl, int ru, int cl, int cu, CacheBlock block) {
...
+		int[] srcIx = new int[offsets.length];
+		Array.copy(offsets, 0, srcIx, 0, offsets.length);
+		int[] destIx = new int[offsets.length];
+		for (int l = 0; l < block.getLength(); l++) {
+			block.set(destIx, get(srcIx));
+			int i = block.getNumDims() - 1;
```
mboehm7: As the input and output tensor block will be of same type shouldn't this implementation be simple a systematic loop over dimensions and array copies in the inner loop?

mboehm7: If you want to leave it like this for now, please put a TODO perf on top.

src/main/java/org/tugraz/sysds/runtime/data/TensorIndexes.java
```diff
@@ -0,0 +1,185 @@
...
+		return sb.toString();
+	}
+
+	////////////////////////////////////////////////////
+	// implementation of Writable read/write
+
+	@Override
+	public void readFields(DataInput in)
+			throws IOException {
+		// TODO readFields
```
mboehm7: Please remove unnecessary TODOs

src/main/java/org/tugraz/sysds/runtime/data/TensorIndexes.java
```diff
@@ -0,0 +1,185 @@
...
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.Arrays;
+
+/**
+ * This represent the indexes to the blocks of the tensor.
+ * Please note that these indexes are 1-based, whereas the data in the block are zero-based (as they are double arrays).
+ */
+public class TensorIndexes implements WritableComparable<TensorIndexes>, RawComparator<TensorIndexes>, Externalizable {
```
mboehm7: As we no longer support Hadoop MapReduce, we can drop the RawComparator implementation here.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/RandSPInstruction.java
```diff
@@ -699,7 +859,84 @@ public GenerateRandomBlock(long rlen, long clen, int brlen, int bclen, double sp
...
+		{
+			//compute local block size:
+			TensorIndexes ix = kv._1();
+			// TODO: accurate block size computation
+			int[] blockDims = new int[_dims.length];
+			blockDims[0] = UtilFunctions.computeBlockSize(_dims[0], ix.getIndex(0), _blen[0]);
+			int clen = 1;
+			for (int i = 1; i < _dims.length; i++) {
+				blockDims[i] = UtilFunctions.computeBlockSize(_dims[i], ix.getIndex(i), _blen[i]);
+				clen *= blockDims[i];
```
mboehm7: Use UtilFunctions.prod instead

src/main/java/org/tugraz/sysds/runtime/instructions/spark/RandSPInstruction.java
```diff
@@ -699,7 +859,84 @@ public GenerateRandomBlock(long rlen, long clen, int brlen, int bclen, double sp
...
+			tb.allocateDenseBlock();
+			if (!_min.equals(_max)) {
+				if (_vt == ValueType.STRING) {
+					throw new DMLRuntimeException("Random string data can not be generated for tensors.");
+				}
+				MatrixBlock blk = new MatrixBlock();
+				RandomMatrixGenerator rgen = LibMatrixDatagen
+						.createRandomMatrixGenerator(_pdf, blockDims[0], clen, blockDims[0], clen, _sparsity,
+								Double.parseDouble(_min), Double.parseDouble(_max), _pdfParams);
+				blk.randOperationsInPlace(rgen, null, seed);
```
mboehm7, line 909: There is something with the formatting off here (too many tabs?)

src/main/java/org/tugraz/sysds/runtime/instructions/spark/AggregateUnarySPInstruction.java
```diff
@@ -109,28 +121,84 @@ public void processInstruction( ExecutionContext ec ) {
...
+
+		// TODO: filter input blocks for trace
+
+		//execute unary aggregate operation
+		AggregateUnaryOperator auop = (AggregateUnaryOperator)_optr;
+		AggregateOperator aggop = _aop;
+
+		// TODO: aggregate
+		/*
+		//perform aggregation if necessary and put output into symbol table
```
mboehm7, line 159: Why is this code completely commented? Wasn't this the main goal of this PR and everything serves to make this instruction work? Anyway, I think it's a good idea to split the changes to DataCharacteristics off, so having this in a separate PR is fine but it would be good to make this work before we move on to the next prototype.

mboehm7 (commented):
Thanks @kev-inn for your effort making these major infrastructure changes. Please address the comments as far as you can, and let's move any major things into separate PRs as this is already getting too large and thus, hard to review.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/AggregateUnarySPInstruction.java
kev-inn: Of course, that was a mistake on my part, I will fix this ASAP

src/main/java/org/tugraz/sysds/runtime/controlprogram/caching/TensorObject.java
kev-inn: I don't quite understand how I could make use of that in this case? This method should update the internal `_metaData`s `DataCharacteristics` with the actual dims and nnz of the `TensorObject`, of which we do not have an accurate `DataCharacteristics` yet, or am I overlooking something/mistake the purpose of the method? Actually `set(DataCharacteristics)` already exists by the way, I just do not know how I could make use of this here.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/RandSPInstruction.java
kev-inn: I fixed it, but I actually did this on purpose since I use two tabs for each newline if I went one level deeper (I will do this differently now).
Basically in this case the first two tabs indent to show that `.creaetRandomMatrixGenerator` is called here and on the next line I have 4 tabs to show that `Double.parseDouble(_min),...` is a continuation of the parameters for `.createRandomMatrixGenerator`, so to make a visual distinction between a method call on LibMatrixDatagen and a method parameter I went for 4 indents.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
kev-inn: Is suppressing the warning fine? The Matrix implementation seems to do that.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
kev-inn: It is correctly handling the boundary blocks, `maxDim` is just not a fitting name, `maxDim` is actually the correct size and not the max size.

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
kev-inn: Should I remove them from the code not written by me?

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
kev-inn: I added a TODO and will improve it later.

kev-inn: Yesterday we talked about the `getNextIndex` procedure, I went ahead and compared it to using modulo for each dimension, this leads to about 12x worse performance for calculating the next index. Instead of 8.22ms we get 103.97ms for calculating all valid indexes for the dimensions `{8, 8, 8, 8, 8, 8, 8}`.

mboehm7:
LGTM - thanks @kev-inn for resolving the open things. I only fixed a couple of warnings and formatting issues.

PR #26, opened 2019-08-06
Initial version of lineage cache with statistics.

phaniarnab:


mboehm7:
LGTM - thanks @phaniarnab, the extension by eviction and statistics is very useful. I just marked the applicability check as a FIXME and reduced the data size of the perf test as we should try to keep the running time of individual tests reasonable. Also, in the future, please try to incorporate the review comments on the original PR.

PR #27, opened 2019-08-07
Replace kahan sum with normal sum for tensor aggregation (uak+ instruction)

kev-inn:
This PR additionally contains small bug fix for `DenseBlockString` and
adds a specific `set(DenseBlock)` implementation for `DenseBlockLString`.


mboehm7:
LGTM - thanks @kev-inn. I added a few TODOs back (e.g., we want to exploit nnz meta data and not recompute the nnz everytime).

PR #28, opened 2019-08-08
Heterogeneous tensors

kev-inn:
This adds a class to support tensors consisting of different datatypes.


kev-inn:
I will resolve the conflicts tomorrow.

kev-inn:
This PR should now be ready for a review.

src/main/java/org/tugraz/sysds/api/DMLScript.java
```diff
@@ -83,7 +83,6 @@
 {
 	private static ExecMode   EXEC_MODE          = DMLOptions.defaultOptions.execMode;    // the execution mode
 	public static boolean     STATISTICS          = DMLOptions.defaultOptions.stats;       // whether to print statistics
-	public static boolean     FINEGRAINED_STATISTICS  = false;                             // whether to print fine-grained statistics
```
mboehm7, line 86: Please don't push other changes from master to your PR unless you squash everything together as this makes it harder to review the actual changes.

src/main/java/org/tugraz/sysds/runtime/controlprogram/caching/TensorObject.java
```diff
@@ -38,7 +38,7 @@
 
 import java.io.IOException;
 
-public class TensorObject extends CacheableData<TensorBlock>
+public class TensorObject extends CacheableData<HomogTensor>
```
mboehm7: No, the tensor object should be using TensorBlock (i.e., the abstract class above BasisTensor and DataTensor) - otherwise this entire class hierarchy has no use.

src/main/java/org/tugraz/sysds/runtime/controlprogram/caching/TensorObject.java
```diff
@@ -105,12 +101,12 @@ public long getNnz() {
 	}
 
 	@Override
-	protected TensorBlock readBlobFromCache(String fname) throws IOException {
```
mboehm7: For all modified method signatures, please see above.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -317,19 +317,19 @@ public static boolean isLocalMaster() {
...
+	public JavaPairRDD<TensorIndexes, HomogTensor> getBinaryTensorBlockRDDHandleForVariable(String varname ) {
 		TensorObject to = getTensorObject(varname);
-		return (JavaPairRDD<TensorIndexes, TensorBlock>)
+		return (JavaPairRDD<TensorIndexes, HomogTensor>)
 				getRDDHandleForTensorObject(to, InputInfo.BinaryTensorBlockInputInfo, -1, true);
 	}
 
 	@SuppressWarnings("unchecked")
-	public JavaPairRDD<TensorIndexes, TensorBlock> getBinaryTensorBlockRDDHandleForVariable(String varname, int numParts, boolean inclEmpty ) {
+	public JavaPairRDD<TensorIndexes, HomogTensor> getBinaryTensorBlockRDDHandleForVariable(String varname, int numParts, boolean inclEmpty ) {
```
mboehm7: Again, this should return TensorBlock as otherwise we would have to introduce the same thing for heterogeneous tensors.

src/main/java/org/tugraz/sysds/runtime/data/HeterogTensor.java
```diff
@@ -0,0 +1,383 @@
...
+import org.tugraz.sysds.common.Types.ValueType;
+import org.tugraz.sysds.runtime.DMLRuntimeException;
+import org.tugraz.sysds.runtime.controlprogram.caching.CacheBlock;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class HeterogTensor extends TensorBlock {
```
mboehm7: Please rename to DataTensor

src/main/java/org/tugraz/sysds/runtime/data/HeterogTensor.java
```diff
@@ -0,0 +1,383 @@
...
+	private HomogTensor[] _colsdata = null;
+	private ValueType[] _schema = null;
```
mboehm7: So you're using a single tensor per column - please add a TODO comment to reconsider using a single tensor for each value type.

src/main/java/org/tugraz/sysds/runtime/data/HeterogTensor.java
```diff
@@ -0,0 +1,383 @@
...
+		if (getNumDims() != 2)
+			throw new DMLRuntimeException("HeterogTensor.get(int,int) dimension mismatch: expected=2 actual=" + getNumDims());
+		return _colsdata[c].get(r, 0);
+	}
+
+	@Override
+	public long getLong(int[] ix) {
+		int[] internalIx = toInternalIx(ix);
+		return _colsdata[ix[1]].getLong(internalIx);
+	}
```
mboehm7: At tensor block level, we could simply return objects and doubles - we only need the type specific methods on DenseBlock because all operations will be implemented against these blocks not against TensorBlock

src/main/java/org/tugraz/sysds/runtime/data/HomogTensor.java
```diff
@@ -0,0 +1,712 @@
+/*
+ * Copyright 2018 Graz University of Technology
```
mboehm7: Update to 2019 license.

src/main/java/org/tugraz/sysds/runtime/data/HomogTensor.java
```diff
@@ -0,0 +1,712 @@
...
+import scala.Array;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+
+public class HomogTensor extends TensorBlock implements Externalizable
```
mboehm7: Please rename to BasicTensor

src/main/java/org/tugraz/sysds/runtime/data/HomogTensor.java
mboehm7: Also did we loose the history again on this rename?

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
```diff
@@ -758,7 +759,28 @@ public static TensorBlock convertToTensorBlock(MatrixBlock mb, ValueType vt) {
 		return ret;
 	}
 
-	public static MatrixBlock[] convertToMatrixBlockPartitions( MatrixBlock mb, boolean colwise ) 
+	public static HeterogTensor convertToHeterogTensor(MatrixBlock mb, ValueType vt) {
```
mboehm7: Maybe change the signature to return a TensorBlock, and pass a boolean argument if this should be a BasicTensorBlock or DataTensorBlock

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
```diff
@@ -955,8 +977,8 @@ public static String toString(TensorBlock mb) {
 	 * @param decimal number of decimal places to print, -1 for default
 	 * @return tensor as a string
 	 */
-	public static String toString(TensorBlock tb, boolean sparse, String separator, String lineseparator,
-			String leftBorder, String rightBorder, int rowsToPrint, int colsToPrint, int decimal){
+	public static String toString(HomogTensor tb, boolean sparse, String separator, String lineseparator,
+	                              String leftBorder, String rightBorder, int rowsToPrint, int colsToPrint, int decimal){
```
mboehm7, line 970: Please never use this Python alignment of arguments, additional lines are supposed to be intended via a single tab.

mboehm7: For long method names + return arguments, this is unreadable and also violates our indentation policy of tabs over spaces.

src/test/java/org/tugraz/sysds/test/component/tensor/HeterogeneousTensorColumnTest.java
```diff
@@ -0,0 +1,233 @@
...
+import org.tugraz.sysds.common.Types.ValueType;
+import org.tugraz.sysds.runtime.data.HeterogTensor;
+import org.tugraz.sysds.runtime.data.HomogTensor;
+
+import java.util.Arrays;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+
+
+public class HeterogeneousTensorColumnTest {
```
mboehm7, line 31: Please rename the tests and test methods according to the renamed class hierarchy

mboehm7 (commented):
Thanks @kev-inn  for refactoring the tensor block abstraction - please rename the implementation classes to BasicTensor/DataTensor or BasicTensorBlock/DataTensorBlock and cleanup the other minor issues, then we can merge it in.

src/main/java/org/tugraz/sysds/api/DMLScript.java
kev-inn: I will try to keep that in mind in the future. I think right now I can not fix this anymore without creating a new PR right?

src/main/java/org/tugraz/sysds/runtime/data/HeterogTensor.java
kev-inn: Could you explain what you mean by "objects and doubles"? Should we have two methods, one for Object (In which I would return either Long, Double, String etc. I guess) and one for returning doubles?

src/main/java/org/tugraz/sysds/runtime/data/HomogTensor.java
kev-inn: I actually don't know how I could keep the history then. I split the rename (`TensorBlock` to `HomogTensor`) and the implementation of the new `TensorBlock` into two commits on purpose so it would not interpret the new `TensorBlock` as the old one, but it seems even that didn't help.

src/main/java/org/tugraz/sysds/runtime/util/DataConverter.java
kev-inn: A single tab or two tabs? I saw both so I am not sure which one I should go for.

src/main/java/org/tugraz/sysds/runtime/data/HeterogTensor.java
kev-inn: Sorry, I completely forgot that we talked about the implementation with only one `BasicTensor` for each datatype.

kev-inn:
Oh I didn't want to do that, I will revert this.

PR #29, opened 2019-08-12
[WIP] New buildin function "slicefinder" 

iulian453:


mboehm7:
Thanks for the initial version @iulian453 - I think it's a good start, so we can merge this in. I deleted the adult.csv dataset and related meta data, the test.dml script, and all changes from other PRs. Furthermore, I simplified the slice-finder Java Test and the script a bit - in general try to keep the scope of local variables as tight as possible and try to avoid unnecessarily convoluted code.

Please, focus now on creating a PR for cleanup. First, this should include a modification of the test such that we automatically check for correct top-k outputs. It's fine if this does not yet always work but it gives people a chance to play with the script and fix it. Second, please comment the DML script and Java test such that people more easily understand what the main parts are trying to do.

PR #30, opened 2019-08-12
Change internal `DataTensor` representation

kev-inn:
Changed the internal representation of columns in `DataTensor`. The interface of `TensorBlock` was also adapted to now instead of implementing a `double`, `long` and `String` getters and setters, there is one each which returns/takes an `Object`.


mboehm7:
LGTM - thanks for the patch @kev-inn. I just made some minor cleanups and replaced some of the type switch statements with `UtilFunctions.objectToDouble()` and `ScalarObjectFactory.createScalar()`.

PR #31, opened 2019-08-20
Add simple binary Spark and CP instructions for tensors

kev-inn:
This should support various binary instructions (+, *, - etc.).
The right hand side tensor (currently only tensor-tensor operations are allowed) does not have to completely match the left hand side dimensions, if the dimension is `1` instead. If an dimension is `1` we reuse this dimension over and over again.
Currently only `repartition join`s are used since size information for tensors is missing in `hops`, which we would need to decide which `SPARK` instruction we want to use (`broadcast` or `repartition`). We could decide between `broadcast` and `repartition` on instruction level, but this would deviate from the matrix implementation.

### DIFF
--- a/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
+++ b/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
@@ -879,6 +879,9 @@ public class BinaryOp extends MultiThreadedHop
 	}
 
 	private static MMBinaryMethod optFindMMBinaryMethodSpark(Hop left, Hop right) {
+		// TODO size information for tensor
+		if (left._dataType == DataType.TENSOR && right._dataType == DataType.TENSOR)
+			return MMBinaryMethod.MR_BINARY_R;
 		long m1_dim1 = left.getDim1();
 		long m1_dim2 = left.getDim2();
 		long m2_dim1 =  right.getDim1();
--- a/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
+++ b/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
@@ -59,12 +59,13 @@ public class BinaryOp extends MultiThreadedHop
 {
 	//we use the full remote memory budget (but reduced by sort buffer), 
 	public static final double APPEND_MEM_MULTIPLIER = 1.0;
-	
+
 	private Hop.OpOp2 op;
 	private boolean outer = false;
 	
 	public static AppendMethod FORCED_APPEND_METHOD = null;
-	
+	public static MMBinaryMethod FORCED_BINARY_METHOD = null;
+
 	public enum AppendMethod { 
 		CP_APPEND, //in-memory general case append (implicitly selected for CP)
 		MR_MAPPEND, //map-only append (rhs must be vector and fit in mapper mem)
@@ -73,7 +74,7 @@ public class BinaryOp extends MultiThreadedHop
 		SP_GAlignedAppend // special case for general case in Spark where left.getCols() % left.getColsPerBlock() == 0
 	}
 	
-	private enum MMBinaryMethod {
+	public enum MMBinaryMethod {
 		CP_BINARY, //(implicitly selected for CP) 
 		MR_BINARY_R, //both mm, mv 
 		MR_BINARY_M, //only mv (mr/spark)
@@ -414,7 +415,7 @@ public class BinaryOp extends MultiThreadedHop
 		} 
 		else 
 		{
-			// Both operands are Matrixes
+			// Both operands are Matrixes or Tensors
 			ExecType et = optFindExecType();
 			boolean isGPUSoftmax = et == ExecType.GPU && op == Hop.OpOp2.DIV && 
 					getInput().get(0) instanceof UnaryOp && getInput().get(1) instanceof AggUnaryOp && 
@@ -433,7 +434,7 @@ public class BinaryOp extends MultiThreadedHop
 				Lop binary = null;
 				
 				boolean isLeftXGt = (getInput().get(0) instanceof BinaryOp) && ((BinaryOp) getInput().get(0)).getOp() == OpOp2.GREATER;
-				Hop potentialZero = isLeftXGt ? ((BinaryOp) getInput().get(0)).getInput().get(1) : null;
+				Hop potentialZero = isLeftXGt ? getInput().get(0).getInput().get(1) : null;
 				
 				boolean isLeftXGt0 = isLeftXGt && potentialZero != null
 					&& HopRewriteUtils.isLiteralOfValue(potentialZero, 0);
@@ -458,6 +459,8 @@ public class BinaryOp extends MultiThreadedHop
 				Hop left = getInput().get(0);
 				Hop right = getInput().get(1);
 				MMBinaryMethod mbin = optFindMMBinaryMethodSpark(left, right);
+				if (FORCED_BINARY_METHOD != null)
+					mbin = FORCED_BINARY_METHOD;
 				
 				Lop  binary = null;
 				if( mbin == MMBinaryMethod.MR_BINARY_UAGG_CHAIN ) {
@@ -467,11 +470,11 @@ public class BinaryOp extends MultiThreadedHop
 							getDataType(), getValueType(), et);
 				}
 				else if (mbin == MMBinaryMethod.MR_BINARY_M) {
-					boolean partitioned = false;
-					boolean isColVector = (right.getDim2()==1 && left.getDim1()==right.getDim1());
-					
+					boolean isColVector = dt1 != DataType.TENSOR && dt2 != DataType.TENSOR &&
+							(right.getDim2() == 1 && left.getDim1() == right.getDim1());
+
 					binary = new BinaryM(left.constructLops(), right.constructLops(),
-							HopsOpOp2LopsB.get(op), getDataType(), getValueType(), et, partitioned, isColVector); 
+							HopsOpOp2LopsB.get(op), getDataType(), getValueType(), et, false, isColVector);
 				}
 				else {
 					binary = new Binary(left.constructLops(), right.constructLops(), 
@@ -487,7 +490,7 @@ public class BinaryOp extends MultiThreadedHop
 
 	@Override
 	public String getOpString() {
-		String s = new String("");
+		String s = "";
 		s += "b(" + HopsOpOp2String.get(op) + ")";
 		return s;
 	}
--- a/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
+++ b/src/main/java/org/tugraz/sysds/hops/BinaryOp.java
@@ -436,8 +436,7 @@ public class BinaryOp extends MultiThreadedHop
 				boolean isLeftXGt = (getInput().get(0) instanceof BinaryOp) && ((BinaryOp) getInput().get(0)).getOp() == OpOp2.GREATER;
 				Hop potentialZero = isLeftXGt ? getInput().get(0).getInput().get(1) : null;
 				
-				boolean isLeftXGt0 = isLeftXGt && potentialZero != null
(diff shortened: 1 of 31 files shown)


src/main/java/org/tugraz/sysds/hops/BinaryOp.java
```diff
@@ -467,11 +470,11 @@ else if(et == ExecType.SPARK)
 							getDataType(), getValueType(), et);
 				}
 				else if (mbin == MMBinaryMethod.MR_BINARY_M) {
-					boolean partitioned = false;
```
mboehm7, line 470: That's a good catch as this was only necessary for map reduce - so maybe remove the argument from the BinaryM lop to get rid of it once and for all.

src/main/java/org/tugraz/sysds/hops/BinaryOp.java
```diff
@@ -487,7 +490,7 @@ else if (mbin == MMBinaryMethod.MR_BINARY_M) {
 
 	@Override
 	public String getOpString() {
-		String s = new String("");
+		String s = "";
 		s += "b(" + HopsOpOp2String.get(op) + ")";
 		return s;
```
mboehm7: Could we maybe fuse this to `return "b(" + HopsOpOp2String.get(op) + ")";`

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -714,11 +714,79 @@ else if(inputInfo2 == InputInfo.BinaryCellInputInfo) {
...
+			int[] blen = new int[dc.getNumDims()];
+			for (int i = 0; i < dc.getNumDims(); i++) {
+				dims[i] = dc.getDim(i);
+				blen[i] = (int) dc.getBlockSize(i);
+			}
+
+			//create partitioned matrix block and release memory consumed by input
+			TensorBlock tb = to.acquireRead();
+			PartitionedBlock<TensorBlock> pmb = new PartitionedBlock<>(tb, dims, blen);
+			to.release();
```
mboehm7: You can say directly acquireReadAndRelease as you anyway hold a strong reference to tb here.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/SparkExecutionContext.java
```diff
@@ -714,11 +714,79 @@ else if(inputInfo2 == InputInfo.BinaryCellInputInfo) {
...
+	}
+
 	public PartitionedBroadcast<MatrixBlock> getBroadcastForVariable(String varname) {
 		MatrixObject mo = getMatrixObject(varname);
 		return getBroadcastForMatrixObject(mo);
 	}
 
+	public PartitionedBroadcast<TensorBlock> getBroadcastForTensorVariable(String varname) {
+		TensorObject to = getTensorObject(varname);
+		return getBroadcastForTensorObject(to);
```
mboehm7: Please fuse this into one line to make it easier to read.

src/main/java/org/tugraz/sysds/runtime/data/BasicTensor.java
```diff
@@ -564,31 +560,54 @@ public BasicTensor aggregateUnaryOperations(AggregateUnaryOperator op, BasicTens
...
 	}
 
+	@Override
+	public TensorBlock binaryOperations(BinaryOperator op, TensorBlock thatValue, TensorBlock result) {
+		if( !LibTensorBincell.isValidDimensionsBinary(this, thatValue) ) {
+			throw new RuntimeException("Block sizes are not matched for binary cell operations");
+		}
+		//prepare result matrix block
+		ValueType vt = resultValueType(_vt, ((BasicTensor) thatValue)._vt);
+		BasicTensor res;
```
mboehm7, line 598: Could you maybe add a check that thatValue is indeed a BasicTensor? It would be great to have a single line similar to `BasicTensor that = checkType(thatValue, true);` which also avoids this potentially unchecked casting.

src/main/java/org/tugraz/sysds/runtime/data/BasicTensor.java
```diff
@@ -648,8 +667,15 @@ public void compactEmptyBlock() {
 
 	@Override
 	public CacheBlock slice(int rl, int ru, int cl, int cu, CacheBlock block) {
-		// TODO Auto-generated method stub
-		return null;
+		BasicTensor bt = (BasicTensor) block;
+		int[] dims = Arrays.copyOf(_dims, _dims.length);
+		dims[0] = ru - rl + 1;
+		dims[1] = cu - cl + 1;
+		bt.reset(dims, false);
```
mboehm7, line 685: Why do we copy it if its immediately overwritten - maybe simplify to `dims = new int[]{ru-rl+1, cu-cl+1}`

src/main/java/org/tugraz/sysds/runtime/data/BasicTensor.java
```diff
@@ -648,8 +667,15 @@ public void compactEmptyBlock() {
 
 	@Override
 	public CacheBlock slice(int rl, int ru, int cl, int cu, CacheBlock block) {
-		// TODO Auto-generated method stub
-		return null;
+		BasicTensor bt = (BasicTensor) block;
```
mboehm7, line 680: Again, please used checked casts with proper exception handling.

src/main/java/org/tugraz/sysds/runtime/data/LibTensorBincell.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * tensor-tensor binary operations
+	 *
+	 * @param m1  input tensor 1
+	 * @param m2  input tensor 2
+	 * @param ret result tensor
+	 * @param op  binary operator
+	 */
+	public static void bincellOp(BasicTensor m1, BasicTensor m2, BasicTensor ret, BinaryOperator op) {
+		// TODO perf (empty, sparse safe, etc.)
```
mboehm7, line 44: Please add a TODO to use separate implementations for matching dims and broadcasting (matrix-vector) which will simplify the implementation to just iterate over all remaining dimensions as done for sum

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -100,7 +104,47 @@ public boolean isEmpty() {
 
 	public abstract double get(int r, int c);
 
+	/**
+	 * Set a cell to the value given as an `Object`. The type is inferred by either the `schema` or `valueType`, depending
+	 * if the `TensorBlock` is a `BasicTensor` or `DataTensor`.
+	 * @param ix indexes in each dimension, starting with 0
+	 * @param v value to set
+	 */
```
mboehm7, line 113: Nice, thanks for keeping documentation on central data structures in mind.

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
```diff
@@ -100,7 +104,47 @@ public boolean isEmpty() {
...
+		if (in1 == ValueType.FP64 || in2 == ValueType.FP64)
+			return ValueType.FP64;
+		if (in1 == ValueType.FP32 || in2 == ValueType.FP32)
+			return ValueType.FP32;
+		if (in1 == ValueType.INT64 || in2 == ValueType.INT64)
+			return ValueType.INT64;
+		if (in1 == ValueType.INT32 || in2 == ValueType.INT32)
+			return ValueType.INT32;
+		if (in1 == ValueType.BOOLEAN || in2 == ValueType.BOOLEAN)
+			return ValueType.INT64;
```
mboehm7, line 153: Why INT64? Ignoring UNKNOWN for know shouldn't it always be boolean & boolean, so return unconditionally return BOOLEAN here? Also please move a check for UNKNOWN with proper exception handling up into the beginning of the method.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/BinarySPInstruction.java
```diff
@@ -160,7 +174,43 @@ protected void processMatrixMatrixBinaryInstruction(ExecutionContext ec)
...
+		updateBinaryTensorOutputDataCharacteristics(sec);
+
+		// Get input RDDs
+		JavaPairRDD<TensorIndexes, TensorBlock> in1 = sec.getBinaryTensorBlockRDDHandleForVariable(input1.getName());
+		JavaPairRDD<TensorIndexes, TensorBlock> in2 = sec.getBinaryTensorBlockRDDHandleForVariable(input2.getName());
+		DataCharacteristics dcOut = sec.getDataCharacteristics(output.getName());
+
+		BinaryOperator bop = (BinaryOperator) _optr;
+
+		// TODO subtensor replication if required
```
mboehm7: yes, please a note on splitting inputs with fewer dimensions into the appropriate blocking scheme.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/data/PartitionedBroadcast.java
```diff
@@ -96,7 +103,27 @@ public T getBlock(int rowIndex, int colIndex) {
...
+			int[] blen = new int[_dc.getNumDims()];
+			for (int i = 0; i < dims.length; i++) {
+				dims[i] = _dc.getDim(i);
+				blen[i] = (int) _dc.getBlockSize(i);
+			}
+			int numPerPart = computeBlocksPerPartition(dims, blen);
+			long l = blen[blen.length - 1];
+			for (int i = blen.length - 2; i >= 0; i--) {
+				l += (ix[i] - 1) * blen[i + 1];
+			}
```
mboehm7: Please make this duplicated index computation a util method somewhere - maybe in `UtilFunctions`.

mboehm7 (commented):
Thanks @kev-inn for these binary tensor operations - this is a great start, please find below some comments, and we can merge it in after we've created our 0.1 release tomorrow.

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
kev-inn: INT64 because for example `sum` should not have boolean as output. But it is true that we might want to overthink this method in itself, maybe once more operations are implemented add an operation parameter?

src/main/java/org/tugraz/sysds/runtime/data/TensorBlock.java
kev-inn: I tried to do it here and there, but didn't yet focus too much on it, since quite a few methods already changed their signature a few times. I will do a PR once we are happy with the general look and document all the central data structures.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/BinarySPInstruction.java
kev-inn: I actually just implemented the functionality itself, so you might want to take a look at that part if the implementation is fine.
EDIT: note functionality was not implemented, instead the replication was implemented

src/main/java/org/tugraz/sysds/runtime/instructions/spark/data/PartitionedBroadcast.java
kev-inn: Are you referring to the calculation of `long l`? Or to the part of getting `dims` and `blen` (which I changed now)?

kev-inn:
The last commit should have resolved the rest of the issues, and complete this PR. Note that I did choose `computeBlockNumber` instead of `computeCellPos` as the name for the `UtilFunction`, because I noticed that the code we used as reference for our interpretation was wrong.

mboehm7:
LGTM - thanks a lot for the patch @kev-inn. I made just minor cleanups (rm warnings/scala imports, clone). However, some tests in binary.tensor/* are still failing - could you please have a look? Thanks.

PR #32, opened 2019-08-20
Kmeans builtin

corepointer:
Opening the PR even though the Spark tests I added are failing (they were not there in the original test case)


mboehm7:
LGTM - thanks @corepointer - I just made some minor syntactic cleanups, removed the codegen aspects from the tests, and fixed tabs over spaces in Java code.

PR #34, opened 2019-08-22
Builtin-reference for steplm and slicefinder

iulian453:


mboehm7:
LGTM - thanks @iulian453 for caring about documentation as well. I just did some minor rewording.

PR #41, opened 2019-09-05
Builtin function lmCV, cross validation with linear regression.

phaniarnab:


scripts/builtin/lmCV.dml
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#-------------------------------------------------------------
+
+m_lmCV = function(Matrix[Double] X, Matrix[Double] y, Integer k, Integer icpt = 0, Double reg = 1e-7) return (Matrix[Double] y_predict)
```
mboehm7: please rename to cvlm or something similar to match the steplm

scripts/builtin/lmCV.dml
```diff
@@ -0,0 +1,47 @@
...
+      testS = 1;
+    else
+      testS = ((i-1) * lim)+1;
+    testE = i * lim;
+
+    testSet = X[testS:testE,];
+    if (i == 1) {
+      trainSet = X[testE+1:M,];
+      trainRes = y[testE+1:M,];
+    }
```
mboehm7: Get rid of this special case at script level - the compiler should simplify it for you.

scripts/builtin/lmCV.dml
```diff
@@ -0,0 +1,47 @@
...
+  M = nrow(X);
+  lim = as.integer(M/k);
+  y_predict = y;
+
+  for (i in 1:k)
+  {
+    if (i == 1)
+      testS = 1;
+    else
+      testS = ((i-1) * lim)+1;
```
mboehm7: Please, use an `testS = ifelse(i==1, 1, ((i-1) * lim)+1)` instead

scripts/builtin/lmCV.dml
```diff
@@ -0,0 +1,47 @@
...
+    if (i == 1) {
+      trainSet = X[testE+1:M,];
+      trainRes = y[testE+1:M,];
+    }
+    else {
+      trainSet = rbind(X[1:testS-1,], X[testE+1:M,]);
+      trainRes = rbind(y[1:testS-1,], y[testE+1:M,]);
+    }
+
+    beta = lm(X=trainSet, y=trainRes, icpt=icpt, reg=reg);
```
mboehm7: Down the road we need to think about using eval instead (and passing arbitrary function pointers) to avoid repeating the same script for every algorithm.

mboehm7: Maybe store and return beta as well?

src/test/java/org/tugraz/sysds/test/functions/builtin/BuiltinLmCVTest.java
```diff
@@ -0,0 +1,71 @@
...
+		
+		int k = 3;
+		proArgs.add("-explain");
+		proArgs.add("-stats");
+		proArgs.add("-args");
+		proArgs.add(input("X"));
+		proArgs.add(input("y"));
+		proArgs.add(String.valueOf(k));
+		proArgs.add(output("B"));
+		programArgs = proArgs.toArray(new String[proArgs.size()]);
```
mboehm7: Maybe directly construct the parameter array instead

src/test/java/org/tugraz/sysds/test/functions/builtin/BuiltinLmCVTest.java
```diff
@@ -0,0 +1,71 @@
...
+		double[][] X = getRandomMatrix(rows, cols, 0, 1, 0.8, -1);
+		double[][] y = getRandomMatrix(rows, 1, 0, 1, 0.8, -1);
+		writeInputMatrixWithMTD("X", X, true);
+		writeInputMatrixWithMTD("y", y, true);
+		
+		runTest(true, EXCEPTION_NOT_EXPECTED, null, -1);
+	}
+	
+
+}
```
mboehm7: Please delete these unnecessary free lines.

mboehm7 (commented):
Thanks for the cv script @phaniarnab - we can merge it once we've done the cleanups below.

scripts/builtin/lmCV.dml
phaniarnab: Didn't understand this comment. Except for the first time, rbind is needed to append two parts to form the training set.

scripts/builtin/lmCV.dml
phaniarnab: Thanks, didn't aware of this. Tried to use this for the next if-else block also, but didn't work, probably because the dimensions of the input matrices of ifelse are not same.

src/test/java/org/tugraz/sysds/test/functions/builtin/BuiltinLmCVTest.java
phaniarnab: This way it is much easier to add and comment out specific options.

scripts/builtin/lmCV.dml
mboehm7: well, in general the else branch would apply to all iterations:
```
trainSet = rbind(X[1:testS-1,], X[testE+1:M,]);
trainRes = rbind(y[1:testS-1,], y[testE+1:M,]);

```
However, our right indexing is not yet able to produce zero row/column matrices (which we would get for X[1:0]). So this is fine for now.

mboehm7:
LGTM - thanks @phaniarnab for the patch and additional cleanup.

PR #44, opened 2019-09-11
Additional reuse options and rbind-tsmm rewrite.

phaniarnab:


mboehm7:
LGTM - thanks for the patch @phaniarnab. I additionally merged the ReuseOptions and CacheType enums as they were kind of redundant.

phaniarnab:
> LGTM - thanks for the patch @phaniarnab. I additionally merged the ReuseOptions and CacheType enums as they were kind of redundant.

Thanks. I kept those separated because CacheType is internal and can grow beyond those 3 types, whereas ReuseOptions is for users.

mboehm7:
good point. Let's keep them as one for now though as we anyway need to expose these types; otherwise they are unusable from outside SystemDS.

phaniarnab:
That is true. Thanks.

mboehm7:
btw, the reason why `append` was used for binary rbind/cbind was to differentiate between binary and nary append during instructions parsing as the latter is not supported for frames. So I'll revert my change and first support lineage tracking for nary rbind/cbind. Once we made the switch to replace frames with our new data tensors, we can clean this up.

PR #45, opened 2019-09-12
Refactoring partial reuse code

phaniarnab:


mboehm7:
LGTM - thanks for the patch @phaniarnab. This is great as it makes adding rewrites easier.

PR #46, opened 2019-09-13
ba+* with rbind and cbind.

phaniarnab:


mboehm7:
LGTM - thanks @phaniarnab. I just fixed the merge issue of the task file and added some code comments. First, it would be awesome if we can have a high-level comment on source and target pattern of the rewrite (I added the target). Second, I put TODOs in the new rewrites - maybe it's a good idea to probe the cache to see if you already have deltaX and deltaY and if yes avoid the right indexing operations.

PR #47, opened 2019-09-18
Remove `istensor` parameter from rand instruction

kev-inn:
Instead of having an explicit parameter to determine if `rand` function should create a matrix or a tensor, we determine this by the presence of either a `dims` parameter or the presence of both `rows` and `cols` parameters.

### DIFF
--- a/src/main/java/org/tugraz/sysds/hops/DataGenOp.java
+++ b/src/main/java/org/tugraz/sysds/hops/DataGenOp.java
@@ -96,10 +96,7 @@ public class DataGenOp extends MultiThreadedHop
 		_op = mthd;
 		
 		//ensure all parameters existing and consistent with data type
-		//TODO remove once this unnecessary parameter is cleaned up
-		if( !inputParameters.containsKey(DataExpression.RAND_TENSOR) )
-			inputParameters.put(DataExpression.RAND_TENSOR, new LiteralOp(false));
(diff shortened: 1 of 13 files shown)


src/main/java/org/tugraz/sysds/hops/DataGenOp.java
```diff
@@ -454,8 +457,8 @@ public boolean hasConstantValue(double val)
 		//sparsity awareness if requires
 		if( ret && val != 0 ) {
 			Hop sparsity = getInput().get(_paramIndexMap.get(DataExpression.RAND_SPARSITY)); //sparsity
-			ret &= (sparsity == null || sparsity instanceof LiteralOp 
-				&& HopRewriteUtils.getDoubleValueSafe((LiteralOp)sparsity) == 1);
+			ret = (sparsity == null || sparsity instanceof LiteralOp
+					&& HopRewriteUtils.getDoubleValueSafe((LiteralOp) sparsity) == 1);
```
mboehm7, line 461: This does not respect the previously computed condition if min and max are equal

mboehm7: ok never mind - its covered under the condition.

mboehm7:
LGTM - thanks @kev-inn

PR #50, opened 2019-09-19
Add non zero tracking for tensors

kev-inn:
Track number of non zero elements for tensors.


mboehm7:
LGTM - thanks @kev-inn. In general this is fine, but let's think about a more elegant way of doing this type-dependent nnz tracking.

PR #51, opened 2019-09-19
Elementwise multiplication

phaniarnab:
Elementwise * with cbind and rbind, stat update to capture execution
time for rewrites.


mboehm7:
LGTM - thanks @phaniarnab. I just made some syntactic changes in the `LineageRewriteTest` (camel case) and I agree we should try to consolidate the various cost models soon.

Furthermore, this change seem to cause an existing lineage test to fail, specifically:```
```
FullReuseTest.testLineageTrace2:
org.tugraz.sysds.runtime.DMLRuntimeException: Variable '_Var18' is not a matrix.
```

phaniarnab:
Yes, that is because * can produce scalar output, for now, will cache only if the output is matrix. I fixed that in local, will merge with next PR.

PR #53, opened 2019-09-24
Add sql builtin to read from database

kev-inn:
TODO: tests, spark, file support.

## Usage:
```r
sql(conn, user, password, query)
sql(conn, query)
```
Parameters can also be specified by name:
```r
sql(query="...", conn="...", ...)
```
## Example:
```
X = sql("jdbc:postgresql://127.0.0.1:5432/postgres", "SELECT * FROM table")
print(toString(X))
```


src/main/java/org/tugraz/sysds/runtime/instructions/cp/SqlCPInstruction.java
```diff
@@ -0,0 +1,125 @@
...
+	private void setCell(TensorBlock outBlock, ResultSet resultSet, ValueType valueType, int[] ix) throws SQLException {
+		int sqlCol = ix[1] + 1;
+		switch (valueType) {
+			case FP64: outBlock.set(ix, resultSet.getDouble(sqlCol)); break;
+			case FP32: outBlock.set(ix, resultSet.getFloat(sqlCol)); break;
+			case INT64: outBlock.set(ix, resultSet.getLong(sqlCol)); break;
+			case INT32: outBlock.set(ix, resultSet.getInt(sqlCol)); break;
+			case BOOLEAN: outBlock.set(ix, resultSet.getBoolean(sqlCol)); break;
+			case STRING: outBlock.set(ix, resultSet.getString(sqlCol)); break;
+		}
```
phaniarnab, line 104: Add may be a default case - that will remove the warning too.

src/main/java/org/tugraz/sysds/runtime/instructions/cp/SqlCPInstruction.java
```diff
@@ -0,0 +1,125 @@
...
+			resultSet = statement.executeQuery(query);
+			ResultSetMetaData meta = resultSet.getMetaData();
+			int cols = meta.getColumnCount();
+			ValueType[] schema = getSchemaFromMetaData(meta);
+			
+			int[] dims = {rows, cols};
+			TensorBlock outBlock = new TensorBlock(schema, dims);
+			for (int row = 0; resultSet.next(); row++)
+				for (int i = 0; i < cols; i++)
+					setCell(outBlock, resultSet, schema[i], new int[]{row, i});
```
phaniarnab, line 85: This cell by cell copy may not work the best for bigger relations. Not sure what I am looking for, but we may need a block by block copy (something along the line of System.arraycopy() ?).

kev-inn: Added TODO since I don't know about a better solution atm.

mboehm7:
LGTM - thanks @kev-inn this is an awesome start. The spark integration and performance features can be added in separate PRs later. During the merge, I only added NUMERIC and DECIMAL as additional types that are mapped to FP64.

PR #54, opened 2019-10-03
cbind-groupedagg, no-spill config

phaniarnab:
Lineage capture for groupedaggregate, cache the same with cbind, new
cache config to disable disk spill


mboehm7:
LGTM - thanks @phaniarnab. I only moved the static no-spill config from DMLScript into the lineage config such that it works though all different APIs (incl. JMLC and MLContext) not just command line.

PR #55, opened 2019-10-04
Performance improvement for append-cbind.

phaniarnab:


mboehm7:
LGTM - thanks @phaniarnab this is a great improvement. However, I fixed the column offset logic as otherwise we created incorrect results if one of the input blocks was empty.

PR #58, opened 2019-10-19
Detection of loop dependent operations.

phaniarnab:
Initial version. This is required to unmark operations for lineage caching.


phaniarnab:
Note: I implemented the fist of the solutions we discussed -- loop till convergence -- to avoid special cases and to make the logic robust.

PR #63, opened 2019-11-09
Initial version of rbind/cbind of List of matrices

phaniarnab:
Single list of only matrices
Only local execution


mboehm7:
LGTM - thanks @phaniarnab. Besides some minor cleanups, I had to revert some of your parser changes as they did not allow traditional nary cbinds/rbinds with more than 3 matrices.

Furthermore, please keep in mind to add the literal replacement for list expansion we talked about Tue. Thanks.

PR #64, opened 2019-11-13
CleanUp_DetectSchema

Shafaq-Siddiqi:
1. Java test file moved to respective Frame package
2. String handling changes with ValueType
3. Double overflow handled
4. Sample storage removed
5. DML test file moved to respective Frame package


mboehm7:
LGTM - thanks for the cleanups @Shafaq-Siddiqi. I only resolved the merge conflicts and removed the new character value type, as we would have to support this throughout the system when introduced. Let's discuss this offline.

PR #65, opened 2019-11-21
Parallel Merge Sort

Baunsgaard:
Parallel merge sort.

include:
- Includes updates to project version dependencies
- hops and lops include thread count for sort operator
- parallel mergesort for CP

comments:
- The LibMatrixReorg file could be reorganized since it is becoming large > 2000 lines of code.
- The Sorting function could perhaps be improved in memory usage if using the Matrixes internal arrays instead of creating new ones.
- The performance was almost comparable to Array.sort() on my 8 cores.
I believe that Java's implementation also is multithreaded.

implementation details
- Splits the array into k parts (number of threads) that is sorted in parallel.
-- splitting into a fixed size was found to be less efficient overall for my machine, but should also be supported.
- Merge Each part two at a time in parallel recursively until the entire is sorted.


corepointer:
Hi Sebastian!

Thank you for your contribution! I have a few comments and suggestions below. After some review and a final LGTM (looks good to me) the PR will usually be squashed to a single commit and merged. 

> Parallel merge sort.
> 
> include:
> 
>     * Includes updates to project version dependencies

Are they related to the topic of this PR? If not, then I suggest we keep updates to dependencies separate from source contributions. This might ease bug tracking if package upgrades have side effects.

> 
> comments:
> 
>     * The LibMatrixReorg file could be reorganized since it is becoming large > 2000 lines of code.
> 

Yes, the code base has grown to a substantial size over the years and Matthias mentioned that they (previous group of people working on SystemML) grouped various pieces of code according to how they saw it fit at the time. Reorganizing it all is a big topic and will be discussed coordinate the effort and have everyone's opinion heard I believe.

>     * The Sorting function could perhaps be improved in memory usage if using the Matrixes internal arrays instead of creating new ones.
> 

I'm sure there is always potential to improve things or do them differently. But as a first warm up exercise it is good that you kept it simple :)

>     * The performance was almost comparable to Array.sort() on my 8 cores.

How big was the input you tested? Maybe max out your RAM ;-)
Also check the JVM options -Xmn<size> -Xms<size> -Xmx<size> 

>       I believe that Java's implementation also is multithreaded.
> 

According to the source of java.utils.Arrays.sort(), an single threaded implementation of Quicksort is used. I noticed a java.utils.Arrays.parallelSort() though.

> 
> implementation details
> 
>     * Splits the array into k parts (number of threads) that is sorted in parallel.
>       -- splitting into a fixed size was found to be less efficient overall for my machine, but should also be supported.
> 

I see you changed that in the source to blklen = 10. Did you test/play around with that value? For large input that might create quite some tasks to process.

.gitignore
```diff
@@ -75,3 +80,10 @@ Gemfile.lock
 **/metastore_db
 derby.log
 /systemds/
+
+
+# Autogenerated files
+src/main/java/org/tugraz/sysds/parser/dml/DmlBaseListener.java
+src/main/java/org/tugraz/sysds/parser/dml/DmlLexer.java
+src/main/java/org/tugraz/sysds/parser/dml/DmlListener.java
+src/main/java/org/tugraz/sysds/parser/dml/DmlParser.java
```
corepointer: Please undo ignoring the auto generated files and take note of the comment in lines 48-60 of .gitignore

src/main/java/org/tugraz/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -356,23 +383,73 @@ public static MatrixBlock sort(MatrixBlock in, MatrixBlock out, int[] by, boolea
...
+				ExecutorService pool = CommonThreadPool.get(k);
+				//compute actual transpose and check for errors
+				// sort smaller blocks.
+				ArrayList<SortTask> tasks = new ArrayList<>();
+				int blklen = 10; 
+				
+				//int blklen = PAR_NUMCELL_THRESHOLD_SORT; 
+				//int blklen = (int)(Math.ceil((double)rlen/k));
+				//blklen += (blklen%8 != 0)?8-blklen%8:0;
+				System.out.println(blklen);
```
corepointer: Maybe use the log facility if you want to keep that output. See for example how LibMatrixNative does it.

src/main/java/org/tugraz/sysds/runtime/matrix/data/LibMatrixReorg.java
```diff
@@ -2341,4 +2455,87 @@ public Long call() {
...
+			_start = start;
+			_end = end;
+			_blockSize = blockSize;
+			_indexes = indexes;
+			_values = values;
+		}
+
+		@Override
+		public Long call(){
+			//Node node = new Node();
```
corepointer: That seems to be a remainder of previous code?

corepointer (changes requested):
Thx again for that good first PR! 
Some more requested changes: 
* please undo the commits that remove the generated parser file and refer to the comment on .gitignore on how to get rid of the issue of seeing changes there (I probably had the same issues - if you use Linux the generated files will change because of the different directory separator)
* please review/explain changes in pom.xml
* same for the changes in the paramserv section

.gitignore
Baunsgaard: Cool, did not see that it was handled like that, thanks for the pointer!

src/main/java/org/tugraz/sysds/runtime/matrix/data/LibMatrixReorg.java
Baunsgaard: - It is unintended should be removed.
- Which logging facility and where do I find an example?

src/main/java/org/tugraz/sysds/runtime/matrix/data/LibMatrixReorg.java
Baunsgaard: Yes, I tried to create a linked list, and that performed worse. 
See https://github.com/tugraz-isds/systemds/pull/65/commits/1e3d8f1375fbc6d3058c8bc78ed1a73fde8429ac

Baunsgaard:
Compare times, in latest commit.

On 8 logical core 6700k Intel processor.

The java times are from line 356 in code. 
sorted array is 2^24 double values in dense one-column matrix.

Times include creating and printing the matrice.

|arrays.sort()|arrays.parallelSort()|my.parallel|old|
| --- | --- | --- | --- |
|2.271s|1.416s|2.475s|3.861s|
|2.363s|1.220s|2.502s|3.752s|
|2.362s|1.486s|2.465s|3.794s|
|2.331s|1.485s|2.507s|3.777s|
|2.339s|1.376s|2.568s|3.753s|
|2.395s|1.459s|2.486s|3.799s|
|2.330s|1.464s|2.514s|3.757s|
|2.412s|1.259s|2.572s|3.788s|
|avg | | | 
|2.351s|1.396s|2.511s|3.785s|

DML script: 
```
s = 2^24
v = 2^20
A = sample(v,s,TRUE,1) + 100
A_i = order(target=A, by=1, decreasing=FALSE, index.return=FALSE);
A_ir = matrix(A_i,(s) / 16, 16 );

print("------SIZE-----")
print(toString(s));
print("------val------")
print(toString(t(A_ir), rows=16, cols=16,decimal=0));
```

Baunsgaard:
> * please undo the commits that remove the generated parser file and refer to the comment on .gitignore on how to get rid of the issue of seeing changes there (I probably had the same issues - if you use Linux the generated files will change because of the different directory separator)

Done

 
> * please review/explain changes in pom.xml

Will be another pull request at some point, the specific versions updated to were to enable direct compatibility with versions used by cloud providers, such as AWS. (might try to run some of it there.)

> * same for the changes in the paramserv section

The changes to paramserv, was due to updating versions. This required interrupted exceptions to be handled. This is again removed in this PR.

src/main/java/org/tugraz/sysds/runtime/matrix/data/LibMatrixReorg.java
corepointer: Ok, a better example than LibMatrixNative would be in InterProceduralAnalysis.java lines 81+82,100-106,184+185. 
You define a log object, a log level and an on/off switch and use it like "if(logLevelIsSet) \ printMessage".

corepointer:
Thx for the testing there. It seems that the Java implementation is already quite good. But you improved performance over what was there and that is nice. After all the goal is to get to know the code base and not to deliver a perfect sorting implementation.

corepointer:
> Compare times, in latest commit.
> 
> On 8 logical core 6700k Intel processor.
> 
> The java times are from line 356 in code.
> sorted array is 2^24 double values in dense one-column matrix.
> 
> Times include creating and printing the matrice.
> arrays.sort() 	arrays.parallelSort() 	my.parallel 	old
> 2.271s 	1.416s 	2.475s 	3.861s
> 2.363s 	1.220s 	2.502s 	3.752s
> 2.362s 	1.486s 	2.465s 	3.794s
> 2.331s 	1.485s 	2.507s 	3.777s
> 2.339s 	1.376s 	2.568s 	3.753s
> 2.395s 	1.459s 	2.486s 	3.799s
> 2.330s 	1.464s 	2.514s 	3.757s
> 2.412s 	1.259s 	2.572s 	3.788s
> avg 			
> 2.351s 	1.396s 	2.511s 	3.785s
> 
> DML script:
> 
> ```
> s = 2^24
> v = 2^20
> A = sample(v,s,TRUE,1) + 100
> A_i = order(target=A, by=1, decreasing=FALSE, index.return=FALSE);
> A_ir = matrix(A_i,(s) / 16, 16 );
> 
> print("------SIZE-----")
> print(toString(s));
> print("------val------")
> print(toString(t(A_ir), rows=16, cols=16,decimal=0));
> ```

Thx for the testing there. It seems that the Java implementation is already quite good. But you improved performance over what was there and that is nice. After all the goal is to get to know the code base and not to deliver a perfect sorting implementation.

corepointer:
> > * please undo the commits that remove the generated parser file and refer to the comment on .gitignore on how to get rid of the issue of seeing changes there (I probably had the same issues - if you use Linux the generated files will change because of the different directory separator)
> 
> Done
> 
> > * please review/explain changes in pom.xml
> 
> Will be another pull request at some point, the specific versions updated to were to enable direct compatibility with versions used by cloud providers, such as AWS. (might try to run some of it there.)
> 
> > * same for the changes in the paramserv section
> 
> The changes to paramserv, was due to updating versions. This required interrupted exceptions to be handled. This is again removed in this PR.

I see. Thx for the explaination. Sounds like a good thing to have in for our 0.2 release at the end of the year.

corepointer:
LGTM :)
Will squash and rebase on Monday if there are no further comments/issues.

corepointer (approved):

mboehm7:
Overall this is awesome - thanks @Sebastian-ba  and @corepointer. In a separate commit I'll make a few cleanups as the current PR did not handle parfor scenarios (where we have to reduce the number of threads of necessary - which deserves a general rework how we do this) and had some opportunities for code simplification. 

Furthermore, for performance benchmarking at script level, we usually use loops to have repetitions inside the JVM which allows for JIT compilation and thus shows the true performance not startup performance. When profiling the new operator I found that only a small fraction was spent in the actual sort, while most of the time went into merge and final copy. Accordingly, I now also made the copy multi-threaded (which improved the performance for 100 1Mx10 order ops from 22.9s to 16.3s). However, we have three barriers and this leads to low utilization - so in the future we might want to reconsider a classical partition sort.

PR #66, opened 2019-11-24
Add construction of federated object with `federated`

kev-inn:
Code is still in its initial stages so recommendations are very welcome. I just wanted to get this "working" stage out there so we have something to base future discussions on. Even though this is really the bare minimum to get `FederatedObjects` into SystemDS it still is much code.

Adds a new function `federated` which takes two parameters `addresses` and `ranges`.

* `addresses`
  This is a list of `String`s of the form `[url/ip]:[port]/[filepath on worker]`
* `ranges`
  This is a list of lists, where the innermost list has 2 values which are the dimension indexes. The `k * 2`th dimension list corresponds to the beginning of dims of the `k`th address workers matrix and the `k * 2 + 1`th dimension list corresponds to the end.

Making `l2svm` work has still a few challenges ahead, such as:
* Should we implement a new `l2svm` called `f_l2svm` or use `m_l2svm`
  I see benefits for both `f_l2svm` would let us treat federated completely separate from matrices and which can be positive, but eliminates possible code reuse
* How exactly should we treat caching of `FederatedObject`
  We could just cache the metadata (addresses, ranges and such), but this might be confusing.


kev-inn:
btw I tried to always use arrays for dimensions so the generalization for tensors will be simplified.

src/main/java/org/tugraz/sysds/runtime/instructions/fed/ReadFEDInstruction.java
```diff
@@ -0,0 +1,136 @@
...
+import org.tugraz.sysds.runtime.data.FederatedRange;
+import org.tugraz.sysds.runtime.instructions.InstructionUtils;
+import org.tugraz.sysds.runtime.instructions.cp.CPOperand;
+import org.tugraz.sysds.runtime.instructions.cp.Data;
+import org.tugraz.sysds.runtime.instructions.cp.IntObject;
+import org.tugraz.sysds.runtime.instructions.cp.ListObject;
+import org.tugraz.sysds.runtime.instructions.cp.ScalarObject;
+import org.tugraz.sysds.runtime.instructions.cp.StringObject;
+import org.tugraz.sysds.runtime.meta.MatrixCharacteristics;
+import org.tugraz.sysds.runtime.meta.MetaData;
```
phaniarnab: Please remove these unused imports.

src/main/java/org/tugraz/sysds/runtime/controlprogram/federated/FederatedRequest.java
```diff
@@ -0,0 +1,70 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.tugraz.sysds.runtime.controlprogram.federated;
+
+import org.apache.hadoop.yarn.webapp.view.HtmlPage;
+import org.tugraz.sysds.runtime.controlprogram.caching.TensorObject;
```
phaniarnab: Please remove these unused imports.

src/main/java/org/tugraz/sysds/runtime/controlprogram/federated/FederatedRequest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
```
phaniarnab: Please remove unused import util.collection.

src/main/java/org/tugraz/sysds/lops/Federated.java
```diff
@@ -0,0 +1,62 @@
...
+		sb.append(OPERAND_DELIMITOR);
+		sb.append(_ranges.prepScalarInputOperand(ranges));
+		sb.append(OPERAND_DELIMITOR);
+		sb.append(prepOutputOperand(output));
+		return sb.toString();
+	}
+	
+	@Override
+	public String toString() {
+		// TODO Sql.toString() lop
```
phaniarnab: Typo in TODO comment.

src/main/java/org/tugraz/sysds/common/Types.java
```diff
@@ -38,14 +38,17 @@
 	 * Data types (tensor, matrix, scalar, frame, object, unknown).
 	 */
 	public enum DataType {
-		TENSOR, MATRIX, SCALAR, FRAME, LIST, UNKNOWN;
+		FEDERATED, TENSOR, MATRIX, SCALAR, FRAME, LIST, UNKNOWN;
```
phaniarnab: What does this new type, FEDERATED mean? Only tensors can be federated or anything from the matrix to list to tensor?

src/main/java/org/tugraz/sysds/runtime/controlprogram/federated/FederatedWorker.java
```diff
@@ -0,0 +1,78 @@
...
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.handler.codec.serialization.ClassResolvers;
+import io.netty.handler.codec.serialization.ObjectDecoder;
+import io.netty.handler.codec.serialization.ObjectEncoder;
+import org.tugraz.sysds.runtime.controlprogram.parfor.util.IDSequence;
+import org.tugraz.sysds.runtime.instructions.cp.Data;
+
+import java.lang.ref.Reference;
```
phaniarnab: Please remove this unused import.

src/main/java/org/tugraz/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
```diff
@@ -0,0 +1,228 @@
...
+					catch (DMLRuntimeException exception) {
+						response = new FederatedResponse(FederatedResponse.Type.ERROR,
+								ExceptionUtils.getFullStackTrace(exception));
+					}
+					finally {
+						ctx.writeAndFlush(response);
+					}
+					break;
+				case VECMATMULT:
+					try {
```
phaniarnab: Do we have a test case for this case?

src/main/java/org/tugraz/sysds/common/Types.java
kev-inn: Since we treat `FederatedObject` differently (we want to call different methods than for example for matrices) I decided to add a new Type. Currently `FEDERATED` refers to an `FederatedObject` which is a `CacheableData<MatrixBlock>`, and it will be generalized for `TensorBlocks` at some point. Does this explain the meaning?

src/main/java/org/tugraz/sysds/runtime/controlprogram/federated/FederatedWorkerHandler.java
kev-inn: No I forgot to remove this before committing, this will be part of another PR

src/main/java/org/tugraz/sysds/common/Types.java
kev-inn: Ok I talked to Matthias and I will remove this new type and do it another way, I will write a comment once I have done that.

kev-inn:
Ok, please take another look, I removed the datatype and `FederatedObject` as requested by Matthias and instead added the functionality to `MatrixObject` for now.

phaniarnab (commented):
Thanks. This looks good to me.

phaniarnab (approved):

mboehm7:
LGTM - thanks @kev-inn and @phaniarnab this is a very nice start. Additionally, I changed the netty dependency to 'provided', integrated the cmd line options handling with our DMLOptions (that uses apache cli), and modified the buffer pool integration so that transferred objects are cached and subject to eviction.

Furthermore, in one of the next PRs please change the stdout logging to a proper logger as otherwise the stdout queue might fill up (because nobody is ready) leading to hanging workers.

PR #67, opened 2019-11-25
Gpu cumulative aggregates

corepointer:


corepointer:
Ran timings on KC GPU server:

```
for(i in 1:10)
{
    A = Rand(rows=65536, cols=1024, min=0, max=1, pdf="uniform", sparsity=1.0, seed=i)
    B = cumsum(A);
    #print(toString(B))
    print(toString(B[nrow(B)-10:nrow(B),1:15], decimal=0, sep=" "))
    #print("Seed=" + i + " sum(cumsum)=" + sum(B))    
}
```

Average from sysds stats module after 10 runs in seconds:

10xCPU | 3.969
10xGPU | 0.871

Average time of cumulative all kernels without memory transfer:
9.55 ms

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/ExecutionContext.java
```diff
@@ -367,7 +373,11 @@ public MatrixObject allocateGPUMatrixObject(String varName, long numRows, long n
 			e = e1;
 		}
 		try {
-			dim2 = validateDimensions(mo.getNumColumns(), numCols);
+			// ToDo: fix this hack
+			if(instName.equals("gpu_ucumk+*"))
```
mboehm7: As you said, please get rid of this hack - there should be no opcodes in the execution context.

src/main/java/org/tugraz/sysds/runtime/controlprogram/context/ExecutionContext.java
```diff
@@ -367,7 +373,11 @@ public MatrixObject allocateGPUMatrixObject(String varName, long numRows, long n
...
+				dim2 = validateDimensions(1, numCols);
+			else
+				dim2 = validateDimensions(mo.getNumColumns(), numCols);
 		} catch(DMLRuntimeException e1) {
 			e = e1;
 		}
```
mboehm7: Since you're on it, please fix this hack of dimension validation via exception handling.

mboehm7:
LGTM - thanks @corepointer. I only fixed some warnings (unnecessary imports) and formatting issues (e.g., tabs vs spaces in *cuh files)

PR #71, opened 2019-12-10
Draft slicing implementation

gilgenbergg:
Implementation of basic approach. With termination constraints based on score upper bound and slice size significance. With first version of basic test setup.


scripts/staging/slicing/base/init.py
```diff
@@ -0,0 +1,45 @@
+import pandas as pd
+from sklearn.linear_model import LinearRegression
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import OneHotEncoder
+
+
+def init():
+    file_name = 'salary.csv'
```
mboehm7: It would be good to refactor this code (into functions and classes) such that we can process any input datasets (e.g., passed as a dataframe). Accordingly, I would recommend to remove hard-coded file names and file I/O from the actual preparation and slicing code.

scripts/staging/slicing/base/init.py
```diff
@@ -0,0 +1,45 @@
...
+    x_size = counter
+    # train model on a whole dataset
+    x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=0.3, random_state=0)
+    model = LinearRegression()
+    model.fit(x_train, y_train)
+    f_l2 = 0
+    for i in range(0, len(y_test)):
+        prediction = model.predict(x_test[i].reshape(1, -1))
+        f_l2_sample = (prediction - y_test[i]) ** 2
+        f_l2 = f_l2 + float(f_l2_sample)
```
mboehm7: Maybe this loop could be vectorized into something like `f_l2 = sum((predict(x_test) - y_test)^2)/x_size`

scripts/staging/slicing/base/main_base.py
```diff
@@ -0,0 +1,127 @@
...
+def opt_fun(fi, si, f, x_size):
+    # weight of error function criteria
+    w = 0.7
+    formula = w * (fi/f) + (1 - w) * (si/x_size)
+    return float(formula)
+
+
+def check_for_nonsense(node_i, node_j, cur_lvl):
+    commons = 0
+    nonsense = False
```
mboehm7: A more descriptive function name would be nice ;-)

scripts/staging/slicing/base/main_base.py
```diff
@@ -0,0 +1,127 @@
...
+        tuple_errors.append(fi_l2_sample)
+        fi_l2 = fi_l2 + fi_l2_sample
+    new_node.l2_error = fi_l2 / new_node.size
+    new_node.score = opt_fun(new_node.l2_error, new_node.size, f_l2, x_size)
+    new_node.e_upper = max(tuple_errors)
+    new_node.c_upper = new_node.score
+    first_level.append(new_node)
+    counter = counter + 1
+levels.append(first_level)
+alpha = 4  # size significance coefficient
```
mboehm7: Why not moving this constraint directly into the optimization function?

scripts/staging/slicing/base/main_base.py
```diff
@@ -0,0 +1,127 @@
...
+                    existing_item = all_nodes[new_node.key]
+                    parents_set = set(existing_item.parents)
+                    parents_set.add(node_i)
+                    parents_set.add(node_j)
+                    existing_item.parents = parents_set
+                else:
+                    new_node.s_upper = new_node.calc_s_upper(cur_lvl)
+                    new_node.s_lower = new_node.calc_s_lower(cur_lvl)
+                    new_node.e_upper = new_node.calc_e_upper()
+                    new_node.c_upper = new_node.calc_c_upper()
```
mboehm7: A `verbose` option of this slicing function would be good to allow more efficient debugging by seeing all upper and lower bounds at once.

scripts/staging/slicing/base/node.py
```diff
@@ -0,0 +1,102 @@
...
+        upper_score = (self.e_upper / self.s_lower) / (f_l2 / x_size) + self.s_upper
+        return float(upper_score)
+
+    def make_slice(self):
+        subset = []
+        for row in complete_x:
+            flag = True
+            for feature in self.attributes:
+                if row[1][feature[1]] != 1:
+                    flag = False
```
mboehm7: Try to vectorize this function if possible as this is used in the inner loop. I could imagine something like 
```
I = (rowSums(complete_x == mask) == numFeatures) # mask being a row vector
subset = select(complete_x, I)
```

scripts/staging/slicing/base/node.py
```diff
@@ -0,0 +1,102 @@
...
+            if flag:
+                subset.append(row)
+        return subset
+
+    def make_first_level_slice(self):
```
mboehm7: What is the difference to the function above.

scripts/staging/slicing/base/node.py
```diff
@@ -0,0 +1,102 @@
...
+        return subset
+
+    def calc_s_upper(self, cur_lvl):
+        sizes = []
+        for parent in self.parents:
+            if cur_lvl == 1:
+                sizes.append(parent.size)
+            else:
+                sizes.append(parent.s_upper)
+        return min(sizes)
```
mboehm7: Instead of appending size information and then calling min on the list, you could directly compute the min.

scripts/staging/slicing/base/node.py
```diff
@@ -0,0 +1,102 @@
...
+    def calc_l2_error(self, subset):
+        fi_l2 = 0
+        for i in range(0, len(subset)):
+            fi_l2_sample = (model.predict(subset[i][1].reshape(1, -1)) - complete_y[subset[i][0]][1]) ** 2
+            fi_l2 = fi_l2 + float(fi_l2_sample)
+        fi_l2 = fi_l2 / len(subset)
+        return float(fi_l2)
+
+
+
```
mboehm7: Unnecessary free lines.

scripts/staging/slicing/datasets/insurance.csv
```diff
@@ -0,0 +1,1339 @@
+age,sex,bmi,children,smoker,region,charges
```
mboehm7: We usually try not to include these datasets, as the potentially have licenses that forbid redistribution. Maybe you could automatically fetch these datasets in your tests?

mboehm7 (commented):
Thanks @gilgenbergg for the patch and welcome as a contributor to SystemDS. Overall, this is a very good start. There are a couple of minor things to address (some of them nice to have). Once this is done, we can merge it in.

scripts/staging/slicing/base/init.py
gilgenbergg: In "tests" folder I added the example of current setup for slicer processing which is now gets constants and prepared initial dataset via arguments.

scripts/staging/slicing/base/node.py
gilgenbergg: Yes, haven't noticed, it differed on the first stages of coding, later method above became more universal but this one below was forgotten to be deleted, sorry for that.

mboehm7:
LGTM - I just added the missing license headers.

PR #72, opened 2019-12-11
Implement federated MV/VM-Multiplication for L2SVM

kev-inn:
Provides a basic implementation for MV/VM-federated operations, which allows us to execute the l2svm builtin.

### DIFF
--- a/src/main/java/org/tugraz/sysds/hops/DataOp.java
+++ b/src/main/java/org/tugraz/sysds/hops/DataOp.java
@@ -113,7 +113,7 @@ public class DataOp extends Hop
 	 * @param dop data operator type
 	 * @param params input parameters
 	 */
-	public DataOp(String l, DataType dt, ValueType vt, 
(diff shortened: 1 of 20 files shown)


kev-inn:
It seems my implementation of async with `CompletableFutures` is faulty, I will have a look into that.

kev-inn:
Btw. the federated instructions for will currently not be shown in the explain plan because the CP instructions are replaced very late when we run the script.

kev-inn:
There is still a problem with some configurations (the configuration file in `conf` causes a request of the whole data for some reason)

mboehm7:
LGTM - thanks @kev-inn that's a great start. Note that I refactored the code in order to move the operations as much as possible out of ProgramBlock and MatrixObject, which is too central to SystemDS for this. I also modified the caching logic to avoid redundant transfers and proper state management. In the future, we should propagate a federated flag during parsing, directly compile federated instructions if we can and only used the `checkAndReplace` if there were dynamic assignments the compiler could not reason about.

PR #73, opened 2019-12-12
Builtin Multinomial Logistic Regression using Trust Region methods

Shafaq-Siddiqi:


src/main/java/org/tugraz/sysds/common/Builtins.java
```diff
@@ -106,13 +105,15 @@
 	LSTM_BACKWARD("lstm_backward", false, ReturnType.MULTI_RETURN),
 	LU("lu", false, ReturnType.MULTI_RETURN),
 	MEAN("mean", "avg", false),
+	MICE_LM("mice_lm", true),
```
Shafaq-Siddiqi, line 108: Two modifications are performed in this file
1. declaration of MultiLogReg builtin
2.  ordering of my previous builtins

mboehm7:
LGTM - thanks @Shafaq-Siddiqi. I only made some minor changes to create consistency of parameter names and defaults across different builtin algorithms.

PR #81, opened 2019-12-23
Lineage tracing for functions. Initial version.

phaniarnab:
- This is an initial version of lineage tracing for functions.
- Need to bring changes for nested functions, cache and reuse
- Need to fix decompression code
- Still thinking about how to get the lineageitem for a function before execution to compare against cached results


phaniarnab:
Shelving this PR for now.
Not to be confused with the other function lineage PR (#84)

PR #82, opened 2019-12-23
Add python wrapper module

kev-inn:
This adds a python wrapper module, with large parts copied from the old systemml solution. Note that currently it does not have all the necessary files for packages (e.g. `setup.py` and `LICENSE`), because there is a lot of metadata, like e-mails and website links, where I do not know what we want to specify.

I made some minor cleanup of the systemml solution and removed the estimators. I will continue to remove parts and adapt other parts, because some features are no longer supported in systemds (like python-like dml), or changed their function signatures (like `var(X, axis=0/1)` being changed to two functions `colVars` and `rowVars`).

I added a dedicated `l2svm` function, but we have to discuss where those script wrappers should be placed.

FYI: I ran the tests with `PYSPARK_PYTHON=python2 spark-submit --master local --driver-class-path [SYSTEMDS-JAR-PATH] [TEST-PYTHON-SCRIPT]`


kev-inn:
I will close this for now, because further discussions lead us to the decision to do some more in depth restructuring of systemml's implementation.

PR #84, opened 2020-01-02
Lineage cache and reuse of function results

phaniarnab:


mboehm7:
LGTM - thanks @phaniarnab. As a follow-up I would recommend to fix the handling of non-deterministic functions by simply not caching outputs that have certain operations (rand, sample, etc) in the lineage obtained while executing the function.

PR #85, opened 2020-01-14
Set intersection operation using tables for numeric data

Shafaq-Siddiqi:


src/test/java/org/tugraz/sysds/test/functions/builtin/BuiltinIntersectionTest.java
```diff
@@ -0,0 +1,120 @@
...
+			loadTestConfiguration(getTestConfiguration(TEST_NAME));
+			String HOME = SCRIPT_DIR + TEST_DIR;
+			fullDMLScriptName = HOME + TEST_NAME + ".dml";
+			programArgs = new String[]{ "-args", input("X"), input("Y"), output("C"), output("X")};
+
+			//generate actual datasets
+			writeInputMatrixWithMTD("X", X, true);
+			writeInputMatrixWithMTD("Y", Y, true);
+
+			runTest(true, false, null, -1);
```
mboehm7, line 114: Please, add a comparison with R here (can be done in a subsequent PR)

scripts/builtin/intersection.dml
```diff
@@ -0,0 +1,57 @@
...
+# intersectionSet           Double    ---         intersection matrix, set of intersecting items
+# intersectCount            Integer               Number of intersecting  items
+
+m_intersection = function(Matrix[Double] X, Matrix[Double] Y)
+  return(Matrix[Double] intersectionSet, Integer intersectCount)
+{
+  X = (table(X, 1) != 0 ) 
+  X = X * seq(1, nrow(X))
+  Y = (table(Y, 1) != 0 ) 
+  Y = Y * seq(1, nrow(Y))
```
mboehm7, line 48: the two multiplications with seq are unnecessary here as they are only used to create an indicator vector

scripts/builtin/intersection.dml
```diff
@@ -0,0 +1,57 @@
...
+# ---------------------------------------------------------------------------------------------
+ 
+#Output(s)
+# ---------------------------------------------------------------------------------------------
+# NAME                      TYPE    DEFAULT      MEANING
+# ---------------------------------------------------------------------------------------------
+# intersectionSet           Double    ---         intersection matrix, set of intersecting items
+# intersectCount            Integer               Number of intersecting  items
+
+m_intersection = function(Matrix[Double] X, Matrix[Double] Y)
```
mboehm7, line 42: please make it m_intersect for consistency with R

scripts/builtin/intersection.dml
```diff
@@ -0,0 +1,57 @@
...
+  return(Matrix[Double] intersectionSet, Integer intersectCount)
+{
+  X = (table(X, 1) != 0 ) 
+  X = X * seq(1, nrow(X))
+  Y = (table(Y, 1) != 0 ) 
+  Y = Y * seq(1, nrow(Y))
+  
+  n = ifelse(nrow(X) < nrow(Y), nrow(X), nrow(Y))
```
mboehm7, line 50: min(nrow(X), nrow(Y)) would be simpler here and is already supported by our size inference engine (which is necessary to compile the subsequent operations into CP)

scripts/builtin/intersection.dml
```diff
@@ -0,0 +1,57 @@
...
+  R = (X[1: n,] * Y[1:n,]) !=0
```
mboehm7, line 51: the !=0 can be dropped as we're now multiplying two 0/1 vectors.

scripts/builtin/intersection.dml
```diff
@@ -0,0 +1,57 @@
...
+  R = R* seq(1,n)
+  R = removeEmpty(target=R, margin="rows")
```
mboehm7, line 53: by feeding the original R into the optional select argument of removeEmpty we can avoid an unnecessary pass.

mboehm7 (commented):
LGTM - I directly made a few teaks of the intersect function. Please finalize it by adding the missing R test comparison. Thanks.

PR #87, opened 2020-01-24
[DIA] Add sum,colSum,rowSum for federated matrices

kev-inn:
Additionally contains some fixes for federated construction and l2svm.

mv/vm (l2svm) does not yet work for execution mode SPARK, while the aggregations and federated construction is expected to work.


kev-inn:
Federated rbind and cbind is also already finished, once this PR is reviewed and I fixed the problems I can immediately create the new PR (since the other depends on some changes done in this PR).

mboehm7:
LGTM - thanks @kev-inn, that's a great start. I simplified the correction handling, fixed the formatting, and consolidated the new LibFederatedAgg into the existing package. 

As a follow-up, it would be good to use for these federated aggregations, the same existing block operations we already use for distributed operations to make them consistent (results and performance).

PR #88, opened 2020-01-26
Functional Dependency - Initial commit

Shafaq-Siddiqi:


mboehm7:
LGTM - thanks @Shafaq-Siddiqi for the initial version. I renamed it, and changed the outer parfor to for as it had a loop-carried dependency.

mboehm7:
After looking over the code a bit more, you were right, there was no parfor dependency as you had separate target matrices.

PR #89, opened 2020-01-27
Outliers detection using standard deviation 

Shafaq-Siddiqi:
The repairs are performed using and repair using row deletion, mean and median imputation.


mboehm7:
LGTM - thanks @Shafaq-Siddiqi. I made some minor cleanups and fixes (e.g., removal of outer). One thing I didn't understand yet, is why the check with sumPrevious and sumNext is still necessary. Could you please have a look when you get a chance?

PR #91, opened 2020-01-28
Outlier detection using IQR - Initial commit

Shafaq-Siddiqi:
Outlier repairs by row deletion, mean and median imputation.


mboehm7:
LGTM - thanks @Shafaq-Siddiqi for the patch and your initiative here. Let's build a library of these primitives and then run some experiments to understand the how much optimal parameters and orders of these primitives matters.

PR #94, opened 2020-02-06
[DIA-22] Second Part - Lineage Remote Parfor Spark

bnyra:
Implement lineage tracing for Remote Parfor Spark

 - Add serialization and de-serialization of lineage in execution context
 - Implement lineage in RemoteParForJobReturn
 - Adapt merging of lineage from different execution contexts
 - Add testcases


bnyra:
@mboehm7, unfortunately I am only available till Saturday 8th February to incorporate feedback. Otherwise I could do so at the end of February.

Best, Benjamin

mboehm7:
LGTM - thanks for the patch @bnyra. There were some issues, but since you're on vacation I decided to fix them and merge it in directly. This completes the DIA project. 

First, we now propagate the lineage flag to remote workers (didn't show up in the tests because they work in a single JVM). Second, we now use a lean way (outside the scope of serialized programs) of serializing and passing the input lineage via task closures. Third, we write the output lineage to HDFS instead of passing it in the output RDD via string concatenation. Fourth, the lineage handling is now conditional as the existing parfor tests without lineage were failing. Fifth and finally, there were some minor formatting issues.

PR #97, opened 2020-02-07
Codegen outer fix index out of bounds with self transposed input

corepointer:


mboehm7:
LGTM - thanks @corepointer. I just made some minor cleanups in the tests. However, the test `testCodegenOuterProd11NoRewrite` is failing and I still don't understand why we need the tsmm flag at runtime, couldn't we find some more generic way of handling it? Could you please have a look into these issues in a subsequent PR?

PR #102, opened 2020-02-11
Cache and reuse statement block results

phaniarnab:


mboehm7:
LGTM - thanks for the patch @phaniarnab. I simplified the code a bit, moved some of the new methods into `LineageItemUtils`, and their invocation in the lineage codepath (out of the non-lineage code-path to ensure a valid comparison between the two modes).

PR #103, opened 2020-02-15
Support of classification analysis

gilgenbergg:
- Calculation of subset's loss (both for regression and classification models) has been updated
- Additional examples of usage were provided


mboehm7:
LGTM - thanks for the patch @gilgenbergg.

PR #105, opened 2020-02-17
Additional slices enumeration approach (union)

gilgenbergg:
In test files option ("union" or "join") is checked in condition according to which specific process method of <slicer> or <union_slicer> is called.

Differences with original slicer module are minimal:
1) Enumeration approach in union_slicer is based on DPsize algorithm (combining all the nodes of previous layers with each other according to number of attributes on current level) checking that no common features have been occurred in forming slices.
2) Additional bounds initialization was required on the first level.


mboehm7:
LGTM - thanks for the patch @gilgenbergg. During the merge I came across the following question: would it make sense for the join-based enumeration approach to also update the bounds in case of existing combinations (e.g., when we created ABCD from ABC and BCD, we could update its bounds when we encounter ABD+BCD)? Let's discuss tomorrow.

PR #114, opened 2020-02-29
Python Wrapper for SystemDS

kev-inn:
We can create matrices from files or numpy. Basic arithmetic is implemented and tested, as well as some aggregations.

Federated works, but I have not added tests because we have to start a worker and we might want to discuss how we distribute the systemds jar. Currently either a standalone jar or our SystemDS jar + classes and libs has to placed in the directory `systemds-java` for the implementation to work, or the java jmlc gateway is manually started and the method `set_aut_start_jmlc(False)` is called.

I did not yet have a detailed look into packaging and releasing on pip, but it seems easy with `setuputils`.

Currently on top of todo list are test cases for federated.

For comparisons like `<`, `>=` etc. we also have to talk about either giving a `DAGNode` back on which `.compute()` has to be called, or calling `.compute()` for both parts and comparing the results. Further discussions which actions except `.compute()` actually start the execution in SystemDS is necessary.


kev-inn:
The failing test case (`Build Failed`) works locally, I will take a look into it.

@Baunsgaard recommended rewriting the test cases to check against DML-Script we expect to be created and not execute it with JMLC Gateway server, therefore we would not have to build the SystemDS jar for the test cases.

Baunsgaard:
The build fails because of one of my tests, it should be fixed if you merge with master again.

src/main/java/org/tugraz/sysds/pythonapi/pythonDMLScript.java
```diff
@@ -0,0 +1,36 @@
...
+	public static void main(String[] args) {
+		GatewayServer gatewayServer = new GatewayServer(new pythonDMLScript());
+		gatewayServer.start();
+		System.out.println("Gateway Server Started");
+		// copied from: https://stackoverflow.com/questions/43282447/py4j-launch-gateway-not-connecting-properly
+		if (args.length == 1 && args[0].equals("--die-on-exit")) {
+			try {
+				BufferedReader stdin = new BufferedReader(new InputStreamReader(System.in));
+				stdin.readLine();
+				System.exit(0);
```
Baunsgaard: If you split this up, such that you make a public function that returns a boolean for if the program fails or passes, then the testing is easier. If you use System exit 0 and 1 then it also will exit the testing.

So i would suggest moving this main methods logic to another function, and then do the System exit here in the main function, while returning a Boolean from the new function.

src/main/java/org/tugraz/sysds/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -103,11 +103,19 @@
 	}
 
 	public static MatrixBlock convertPy4JArrayToMB(byte [] data, long rlen, long clen) {
-		return convertPy4JArrayToMB(data, (int)rlen, (int)clen, false);
+		return convertPy4JArrayToMB(data, (int)rlen, (int)clen, false, 2);
```
Baunsgaard: why cast? can the inputs not be Ints per default?

src/main/python/setup.py
```diff
@@ -0,0 +1,97 @@
...
+    These high-level scripts are compiled into hybrid execution plans of local, in-memory CPU and GPU operations, 
+    as well as distributed operations on Apache Spark. In contrast to existing systems 
+    - that either provide homogeneous tensors or 2D Datasets - and in order to serve the entire
+     data science lifecycle, the underlying data model are DataTensors, i.e.,
+     tensors (multi-dimensional arrays) whose first dimension may have a heterogeneous and nested schema.
+
+    ''',
+    url='https://github.com/tugraz-isds/systemds',
+    author='SystemDS',
+    author_email='dev@systemds.apache.org',
```
Baunsgaard: is / will this be correct?

src/main/python/setup.py
```diff
@@ -0,0 +1,97 @@
...
+setup(
+    name=ARTIFACT_NAME,
+    version=ARTIFACT_VERSION_SHORT,
+    description='Apache SystemDS is a distributed and declarative machine learning platform.',
+    long_description='''
+    
+    SystemDS is a versatile system for the end-to-end data science lifecycle from data integration, 
+    cleaning, and feature engineering, over efficient, local and distributed ML model training, 
+    to deployment and serving. To this end, we aim to provide a stack of declarative languages with
+    R-like syntax for
```
Baunsgaard: Maybe not mention R?

Suggested change:

SystemDS is a versatile system for the end-to-end data science lifecycle from data integration, 
    cleaning, and feature engineering, over efficient, local and distributed ML model training, 
    to deployment and serving.
To facilitate this, bindings from different languages and different system abstractions provide help for:

src/main/python/setup.py
```diff
@@ -0,0 +1,97 @@
...
+    description='Apache SystemDS is a distributed and declarative machine learning platform.',
+    long_description='''
+    
+    SystemDS is a versatile system for the end-to-end data science lifecycle from data integration, 
+    cleaning, and feature engineering, over efficient, local and distributed ML model training, 
+    to deployment and serving. To this end, we aim to provide a stack of declarative languages with
+    R-like syntax for 
+    (1) the different tasks of the data-science lifecycle, and 
+    (2) users with different expertise. 
+    These high-level scripts are compiled into hybrid execution plans of local, in-memory CPU and GPU operations,
```
Baunsgaard: Newline!

src/main/python/systemds/converters.py
```diff
@@ -0,0 +1,46 @@
...
+def numpy_to_matrix_block(jvm: JVMView, np_arr: np.array):
+    assert (np_arr.ndim <= 2)
+    rows = np_arr.shape[0]
+    cols = np_arr.shape[1] if np_arr.ndim == 2 else 1
+    if not isinstance(np_arr, np.ndarray):
+        np_arr = np.asarray(np_arr, dtype=np.float64)
+    # data_type: 0: int, 1: float and 2: double
+    if np_arr.dtype is np.dtype(np.int32):
+        arr = np_arr.ravel().astype(np.int32)
+        data_type = 0
```
Baunsgaard: we might want to consider something else than marking it as a integer for each datatype.
There should be enum types in python as i remember.
The other option is to use the already defined types from NumPY.

src/main/python/systemds/matrix.py
```diff
@@ -0,0 +1,127 @@
...
+    for begin, end in ranges:
+        ranges_str += f'list({",".join(map(str, begin))}), list({",".join(map(str, end))})'
+    ranges_str += ')'
+    named_params = {'addresses': addresses_str, 'ranges': ranges_str}
+    named_params.update(kwargs)
+    return OperationNode('federated', args, named_params)
+
+
+# TODO consider different naming for functions
+def full(shape: Tuple[int, int], value: Union[float, int]) -> OperationNode:
```
Baunsgaard, line 100: call it dense Matrix maybe?

src/main/python/systemds/matrix.py
```diff
@@ -0,0 +1,127 @@
...
+    :param shape: shape (rows and cols) of the matrix TODO tensor
+    :param value: the value to fill all cells with
+    :return: the DAGNode representing this operation
+    """
+    unnamed_input_nodes = [value]
+    named_input_nodes = {'rows': shape[0], 'cols': shape[1]}
+    return OperationNode('matrix', unnamed_input_nodes, named_input_nodes)
+
+
+def seq(start: Union[float, int], stop: Union[float, int] = None, step: Union[float, int] = 1) -> OperationNode:
```
Baunsgaard, line 112: Call it Vector?

src/main/python/systemds/operation_node.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def abs(self) -> OperationNode:
+        """
+        Calculate absolute.
+        :return: `OperationNode` representing operation
+        """
+        return OperationNode('abs', [self])
+
+    def moment(self, moment, weights: DAGNode = None) -> OperationNode:
+        # TODO write tests
```
Baunsgaard, line 185: why is this comment only here?

src/main/java/org/tugraz/sysds/pythonapi/pythonDMLScript.java
```diff
@@ -0,0 +1,36 @@
+package org.tugraz.sysds.pythonapi;
```
Baunsgaard: missing license

src/main/java/org/tugraz/sysds/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
```diff
@@ -103,11 +103,19 @@
 	}
 
 	public static MatrixBlock convertPy4JArrayToMB(byte [] data, long rlen, long clen) {
-		return convertPy4JArrayToMB(data, (int)rlen, (int)clen, false);
```
Baunsgaard, line 106: Update License

src/main/resources/python/project_info.py
```diff
@@ -0,0 +1,27 @@
+#-------------------------------------------------------------
+#
```
Baunsgaard: Update to Modification License

Baunsgaard (commented):
LGTM,  :100: :1st_place_medal: :+1:   
only cosmetic comments

src/main/java/org/tugraz/sysds/runtime/instructions/spark/utils/RDDConverterUtilsExt.java
kev-inn: I removed the long methods for now, because I agree they should never get used by py4j at the moment.

src/main/python/systemds/operation_node.py
kev-inn: because this function is the only one that is still missing tests.

kev-inn:
Btw I specified `author_email='damslab@mlist.tugraz.at'`, I do not know if this is the address we want to be contacted by.

src/main/python/systemds/matrix.py
Baunsgaard: like the numpy seq

kev-inn:
Documentation will be added soon.

src/main/python/NOTICE
```diff
@@ -0,0 +1,8 @@
+SystemDS
+Copyright 2018 Graz University of Technology
```
Baunsgaard: update

src/main/python/pre_setup.py
```diff
@@ -0,0 +1,44 @@
...
+import fnmatch
+
+python_dir = 'systemds'
+java_dir = 'systemds-java'
+java_dir_full_path = os.path.join(python_dir, java_dir)
+if os.path.exists(java_dir_full_path):
+    shutil.rmtree(java_dir_full_path, True)
+os.mkdir(java_dir_full_path)
+root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.getcwd())))
+JAR_FILE_NAMES = ['systemds-*-SNAPSHOT.jar', 'systemds-*.jar', 'systemds-*-SNAPSHOT-extra', 'systemds-*-extra.jar']
```
Baunsgaard, line 36: Maybe not take any SystemDS, but the latest one?

src/main/python/setup.py
```diff
@@ -0,0 +1,100 @@
...
+    print("Could not read project_info.py. Will use default values.", file=sys.stderr)
+    BUILD_DATE_TIME = str(time.strftime("%Y%m%d.%H%M%S"))
+    __project_artifact_id__ = 'systemds'
+    __project_version__ = BUILD_DATE_TIME + '.dev0'
+ARTIFACT_NAME = __project_artifact_id__
+ARTIFACT_VERSION = __project_version__
+ARTIFACT_VERSION_SHORT = ARTIFACT_VERSION.split("-")[0]
+
+numpy_version = '1.8.2'
+py4j_version = '0.10.0'
```
Baunsgaard: You might want to make a requirements.txt also, for easy standard installation.

https://stackoverflow.com/questions/7225900/how-to-install-packages-using-pip-according-to-the-requirements-txt-file-from-a#15593865

Baunsgaard (commented):
minor things for the changes.

src/main/python/pre_setup.py
kev-inn: This would require us to each time we want to distribute a new package to make changes to this file, instead of just compiling with maven and executing `pre_setup.py`, `setup.py` and `post_steup.py`. I would prefer the second option.

src/main/python/setup.py
kev-inn: Setuputils creates the requirements file automatically, that`s what the `setup.py` is for.

src/main/python/NOTICE
kev-inn: should I replace with `[2018-2020]` or `2020`. The `NOTICE` file in the root dir should then also be updated I guess?

Baunsgaard: I think that we can update the notice to say `[2018-2020]` and then write 2020 in these files.

src/main/python/pre_setup.py
Baunsgaard: true, but it would ensure that the interface it compatible with the system. If you don't like it, then do not make the change :)

kev-inn:
Note: the building of the packaging needs to be work in combination with @corepointer 's efforts towards 0.2.0, and the reworked `pom.xml`, therefore we could either accept one of the PR's and merge it into the other one, or merge both and create a whole new PR for packaging (I could delete the packaging files `setup.py`, `pre_setup.py`... for now.)

Baunsgaard:
I'm Merging this into master, with some slight modifications.
While doing so i have problems executing the tests, that all result in resource warnings.

`/usr/lib/python3.7/subprocess.py:883: ResourceWarning: subprocess 32620 is still running`

what was the fix again @kev-inn ?

also we have to talk about supported versions, since trying it with anything older than python 3.7 doesn't work, and as far as i know python 3.6 is the default ATM.

kev-inn:
@Baunsgaard I will look into the warning.

Yes, I depend on 3.7 for https://www.python.org/dev/peps/pep-0563/ . There is an old way of doing the delayed evaluation, I will switch to that.

kev-inn:
@Baunsgaard concerning the warning: It happens because we start a subprocess for the SystemDS execution, but it is hard to determine when we want to close that started process, because we never know that no more SystemDS execution is necessary. I tried to do it `atexit`, but py4j doesn't appreciate that. I will look into it, but if I feel like a `Context` is necessary I might postpone it for now.
Strangely if I execute the scripts in intellij I do not get the warning, I will take a closer look, but focus on other things for now.

PR #122, opened 2020-03-20
Docker

Baunsgaard:
Docker integration! easy to use also!

### DIFF
--- a/.github/action/Dockerfile
+++ b/.github/action/Dockerfile
@@ -1,6 +1,6 @@
 #-------------------------------------------------------------
 #
-# Modifications Copyright 2020 Graz University of Technology
+# Copyright 2020 Graz University of Technology
 #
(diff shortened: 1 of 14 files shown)


Baunsgaard:
WIP,
But you can try it out by reading the docker/README.md.

Only two commands needed for Systemds (sebaba is my docker profile name)

```bash
docker pull sebaba/sysds:0.2
docker run \
  -v $(pwd)/docker/mountFolder:/input \
  --rm sebaba/sysds:0.2
```

Baunsgaard:
Improved startup times in Workflow tests!
Example: 
1m 59s total for Application Tests starting with A
time before 14m 35s

This speedup in execution is true for all Application Tests, and Function Tests in  a scale of 13-15 min each test as a constant speedup.

PR #123, opened 2020-03-21
Expose lineage trace for python-systemDS

phaniarnab:
This patch contains the implementation of exposing lineage trace to python users. Lineage trace can be retrieved by one of the below ways:
1) by passing a lineage = True to compute() function
```R
print(m_sum.compute(verbose = True, lineage = True))
```
2) by calling getLineageTrace() on any intermediate.
```R
print(m_res.getLineageTrace())
```

Note: Unit test setup is not ready yet but the reviewer can execute test_lineage.py and test_l2svm_lineage.py files to display lineage trace.


src/main/python/systemds/matrix/matrix.py
```diff
@@ -65,13 +65,16 @@ def code_line(self, var_name: str, unnamed_input_vars: Sequence[str],
...
         if self._is_numpy():
             if verbose:
                 print('[Numpy Array - No Compilation necessary]')
             return self.np_array
         else:
-            return super().compute(verbose)
+            return super().compute(verbose, lineage)
+
+    def getLineageTrace(self) -> str:
+        return super().getLineageTrace()
```
kev-inn: I think you do not need the lines, because they will anyway be inherited from `OperationNode`

src/main/python/systemds/script_building/script.py
```diff
@@ -79,7 +79,34 @@ def execute(self) -> JavaObject:
...
             for (name, input_node) in self.inputs.items():
                 input_node.pass_python_data_to_prepared_script(gateway.jvm, name, self.prepared_script)
-        return self.prepared_script.executeScript()
+
+            if lineage:
+                connection.setLineage(True)
+
+        ret = self.prepared_script.executeScript()
+        if lineage:
+            print(self.prepared_script.getLineageTrace(self.out_var_name))
```
kev-inn: Maybe we could return this instead so the user can decide if he wants to print it or maybe write it to a file or do something else with it.

src/main/python/systemds/script_building/script.py
```diff
@@ -79,7 +79,34 @@ def execute(self) -> JavaObject:
...
+                                                            _list_to_java_array([self.out_var_name]))
+            for (name, input_node) in self.inputs.items():
+                input_node.pass_python_data_to_prepared_script(gateway.jvm, name, self.prepared_script)
+
+            connection.setLineage(True)
+
+        ret = self.prepared_script.executeScript()
+        lineage = self.prepared_script.getLineageTrace(self.out_var_name)
+        return lineage
+
```
kev-inn, line 109: Could we maybe add this to `.executeScript()`, since the code seems to be the almost the same

src/main/python/systemds/script_building/script.py
kev-inn: That would also allow us to better test it

src/main/python/tests/test_l2svm_lineage.py
```diff
@@ -0,0 +1,29 @@
...
+    np.random.seed(seed)
+    m1 = np.array(np.random.randint(100, size=dims * dims) + 1.01, dtype=np.double)
+    m1.shape = (dims, dims)
+    m2 = np.zeros((dims, 1))
+    for i in range(dims):
+        if np.random.random() > 0.5:
+            m2[i][0] = 1
+    return Matrix(m1), Matrix(m2)
+
+print(getl2svm())
```
kev-inn: Maybe create a class `class TestLineage(unittest.TestCase):` like the other UnitTests do it.

src/main/python/tests/test_lineage.py
```diff
@@ -0,0 +1,10 @@
+from systemds.matrix import full
+m = full((5, 10), 4.20)
+m_res = m * 3.1
+m_sum = m_res.sum()
+
+#print lineage trace of intermediate m_res
+print(m_res.getLineageTrace())
+
+#compute and print lineage trace of the full dag lazily
+print(m_sum.compute(verbose = True, lineage = True))
```
kev-inn: Same as with the other testcase

kev-inn (commented):
LGTM :100: I added some comments you might want to address :smiley:

src/main/python/systemds/matrix/matrix.py
kev-inn: Or you might want to add an if/else for the case that we have a numpy array (which we would not pass to systemds if `.compute()` is called directly on this `Matrix` object)

src/main/python/systemds/script_building/script.py
phaniarnab: It didn't look good to me that compile() is returning the trace, that is why I provided the other separate function getLineageTrace() to get the lineage trace of any intermediates.

src/main/python/systemds/script_building/script.py
phaniarnab: This looks the same because to get the lineage you need to execute the script. But from an user's point, they do different things.

src/main/python/tests/test_l2svm_lineage.py
phaniarnab: yes, I need to do that.

src/main/python/systemds/script_building/script.py
Baunsgaard: one of the strengths of python is that you are able to return multiple things, so in this case you could just return it, and on the receiving end (where the method is called) it can be filtered out.

src/main/python/systemds/script_building/script.py
kev-inn: Will the user ever have direct access to the Script? I think currently Script is a non public object (we might change that in the future).

phaniarnab: This routine is not accessed by the users, they call getLineageTrace which internally calls this. But you are right that as now compute too can return lineage trace, that can be reused. Will do it later.

corepointer (approved):
LGTM & thx for the contribution. Will resolve conflicts and merge.

PR #125, opened 2020-03-22
Python test debug

corepointer:
creating a PR for this to check if all tests work now (tests triggered by PR to master)
will merge myself


kev-inn:
I would prefer some other way to shutdown the server than `shutdown`. Fist of all a check is missing since we only start a subprocess if no SystemDS instance is running yet, secondly if we need a explicit shutdown it makes sense to also have an explicit start. As I have already suggested in my PR it would make sense to use a ContextManager (See https://book.pythontips.com/en/latest/context_managers.html). We should make sure that we are happy with our decision since it will likely stay for longer, I hoped that we find a way to automatically shutdown the java server process on exit of the python program, but a ContextManager would also be a good idea.

PR #133, opened 2020-03-26
[MINOR] Federated Python Docs

Baunsgaard:
Added federated documentation example for python.

:bug: fix in python federated not allowing for multiple federated environments.


bin/README.md
```diff
@@ -19,46 +19,54 @@ limitations under the License.
...
+bin/systemds.sh Univar-Stats.dml -nvargs X=data/haberman.data TYPES=data/types.csv STATS=data/univarOut.mtx CONSOLE_OUTPUT=TRUE
 ```
 
-#### Using Intel MKL native instructions
-To use the MKL acceleration download and install the latest MKL library from [1], 
-set the environment variables with the MKL-provided script `$ compilervars.sh intel64` and set 
+### Using Intel MKL native instructions
+
+To use the MKL acceleration download and install the latest MKL library from [1],
+set the environment variables with the MKL-provided script `$ compilervars.sh intel64` and set
```
Baunsgaard, line 90: Here the [1] points to where?

docker/build.sh
```diff
@@ -30,5 +30,8 @@ docker image build -f docker/sysds.Dockerfile -t sebaba/sysds:0.2 .
 # The second build is for testing systemds. This image installs the R dependencies needed to run the tests.
 docker image build -f docker/testsysds.Dockerfile -t sebaba/testingsysds:0.2 .
 
+# The third build is python docker for systemds. 
+docker image build -f docker/pythonsysds.Dockerfile -t sebaba/pythonsysds:0.2 .
```
Baunsgaard, line 34: python docker image to tests the pip install.

src/main/python/systemds/matrix/matrix.py
```diff
@@ -90,7 +90,8 @@ def federated(addresses: Iterable[str], ranges: Iterable[Tuple[Iterable[int], It
     addresses_str = 'list(' + ','.join(map(lambda s: f'"{s}"', addresses)) + ')'
     ranges_str = 'list('
     for begin, end in ranges:
-        ranges_str += f'list({",".join(map(str, begin))}), list({",".join(map(str, end))})'
+        ranges_str += f'list({",".join(map(str, begin))}), list({",".join(map(str, end))}),'
+    ranges_str = ranges_str[:-1]
```
Baunsgaard, line 94: add , between each range element.

Baunsgaard:
@corepointer  I have changed it such that it is contained inside the systemds.sh script.

Might i come with the suggestion that we switch the ending of that file to nothing?

This would make the commands look nicer.

```bash
systemds WORKER 8001
```

compared to 

```bash
systemds.sh WORKER 8001
```

bin/README.md
corepointer: * Please keep non-changes that your editor makes to yourself. I have a hard time spotting any real changes in that file.  
* The missing reference should lead to https://software.intel.com/mkl

docker/pythonsysds.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openjdk:8-alpine
```
corepointer, line 1: so you stayed with Alpine after all?

src/main/python/docs/source/federated.rst
```diff
@@ -0,0 +1,96 @@
...
+The python SystemDS supports federated execution.
+To enable this, each of the federated environments have to have 
+a running federated worker.
+
+Start Federated worker
+----------------------
+
+To start a federated worker:
+
+  ./bin/systemds.sh WORKER 8001
```
corepointer: If that will also work by unzipping a (future) binary release, then you could omit bin/

src/main/python/docs/source/federated.rst
```diff
@@ -0,0 +1,96 @@
...
+First we need to create some data for our federated worker to use.
+In this example we simply use Numpy to create a ``test.csv`` file.::
+
+  # Import numpy
+  import numpy as np
+  a = np.asarray([[1,2,3],[4,5,6],[7,8,9]])
+  np.savetxt("temp/test.csv", a, delimiter=",")
+
+Currently we also require a metadata file for the federated worker.
+This should be located next to the ``test.csv`` file, containing::
```
corepointer: mention that this metadata file should be called test.csv.mtd or <same-name-plus-extension>.mtd

src/main/python/docs/source/federated.rst
```diff
@@ -0,0 +1,96 @@
...
+  {
+    "data_type": "matrix",
+	"format": "csv",
+	"header": false,
+	"rows":3,
+	"cols":3 
+  }
+
+After creating our data we start a Federated worker instance the 
+federated instructions can be executed.
```
corepointer: Does Federated need the capital F? Leaving out one of the federated would also be a good option. And there's an "on" missing at the end of the sentence (...start an instance... the instructions can be executed *on*). Maybe it's not *on*, but you execute on/in/at an instance. Or you make two sentences where you start in the first and execute in the second one.

src/main/python/docs/source/federated.rst
```diff
@@ -0,0 +1,96 @@
...
+The multiply using federated instructions in python SystemDS is done
+as follows::
+
+  # Import numpy and SystemDS federated
+  import numpy as np
+  from systemds.matrix import federated
+  # Create a federated matrix
+  fed_a = federated(["localhost:8001/temp/test.csv"],[([0,0],[3,3])])
+  print(fed_a.sum().compute())
+  # Result should be 45.
```
corepointer: * You mention multiply and do a sum.
* You should explain the federated() a bit more or point to where it's documented.

src/main/python/docs/source/federated.rst
```diff
@@ -0,0 +1,96 @@
...
+Multiple Federated environments
+-------------------------------
+
+Using the data created from the last example we can simulate
+multiple federated workers by simply starting multiple ones on different ports.
+
+
+| ./bin/systemds.sh WORKER 8001
+| ./bin/systemds.sh WORKER 8002
+| ./bin/systemds.sh WORKER 8003
```
corepointer: What does the pipe ( | ) do here?

corepointer (changes requested):
In general very good. A few minor changes would improve the PR.
Having the federated functionality in the systemds script could have been done without explicitly having to type WORKER, but your solution looks solid and self explanatory :+1: 
If you could fix that missing reference that I forgot to put into the readme file, I will retract my complaint about not seeing the necessary change in that file :wink:

corepointer:
> @corepointer I have changed it such that it is contained inside the systemds.sh script.
> 
> Might i come with the suggestion that we switch the ending of that file to nothing?
> 
> This would make the commands look nicer.
> 
> ```shell
> systemds WORKER 8001
> ```
> 
> compared to
> 
> ```shell
> systemds.sh WORKER 8001
> ```

I agree, that it would look nicer without the file extension. On the other hand I fancy a descriptive file extension. If it's a script, make it look like a script. Again on the other hand (running out of hands here *g*), many projects, especially java based ones, seem to have grown the habit of disguising their startup scripts like real binaries. But that is just my humble opinion and I can live with a rename. Maybe ask for more opinions :point_right:

src/main/python/docs/source/federated.rst
Baunsgaard: sure, ill remove the bin part :)

docker/pythonsysds.Dockerfile
Baunsgaard: yes, i fought through it ... now 500MB image instead of 1-2GB. There is more things that could be done to decrease the size. but not something i will work on now

src/main/python/docs/source/federated.rst
Baunsgaard: makes the documentation force newlines. (i heard you don't like blank newlines.)

